PR #7, opened 2014-01-05
Make Field.Option public so that it is visible outside the package

deafgreatdane:
In order to make the field options usable by the client (such as for mimicking the pulldown choices in our own UI), the Option class needs to be public


russellhoff:
Option constructor remains private and it's unusable. Please @stephenc , could you revert the changes made by @deafgreatdane ?

russellhoff:
I've opened a new pull request #174, please accept it.

PR #20, opened 2014-11-21
Support for Triggers and Targets

artichikin:
This works with the current exception of Zendesk.getTargets(). The zendesk server response is currently inconsistent with it’s other list responses, and PagedIterator currently relies on this behaviour for loop termination.


johnou:
@artichikin nice work on implementing triggers and targets, do you have time to fix the merge conflicts so @stephenc can merge your PR?

src/main/java/org/zendesk/client/v2/model/Trigger.java
```diff
@@ -0,0 +1,248 @@
...
+         final StringBuilder sb = new StringBuilder();
+         sb.append("Action");
+         sb.append("{field=").append(field);
+         sb.append(", value=").append(value);
+         sb.append('}');
+         return sb.toString();
+      }
+
+      // Zendesk sometimes returns JSON with a String instead of String[] so we need a custom deserializer
+      public static class ActionDeserializer extends JsonDeserializer<Action> {
```
johnou: @artichikin we noticed the same thing but didn't manage to send a pull request with our implementation, the downside is it contains a pretty nasty hack for something we were not able to figure out..

```java
/**
 * @author Johno Crawford (johno@sulake.com)
 */
public class ActionDeserializer extends com.fasterxml.jackson.databind.JsonDeserializer<Action> {

    @Override
    public Action deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
        ObjectCodec objectCodec = jsonParser.getCodec();
        JsonNode root = objectCodec.readTree(jsonParser);
        String value = "";
        JsonNode valueNode = root.get("value");
        if (valueNode.isArray()) {
            for (JsonNode node : valueNode) {
                if (!node.asText().startsWith("channel:")) {
                    value = node.asText();
                    break;
                }
            }
        } else {
            value = valueNode.asText();
        }
        return new Action(root.get("field").asText(), value);
    }
}
```

artichikin:
I'm going to do a new pull request with our latest changes, merged up

src/main/java/org/zendesk/client/v2/model/Trigger.java
```diff
@@ -0,0 +1,246 @@
...
+         sb.append(", operator=").append(operator);
+         sb.append(", value=").append(value);
+         sb.append('}');
+         return sb.toString();
+      }
+
+   }
+
+   @JsonDeserialize(using = org.zendesk.client.v2.model.Trigger.Action.ActionDeserializer.class)
+   public static class Action {
```
johnou: I think it makes more sense to pull this class up one level, Action is also used for Macro.

johnou:
@artichikin any luck with the new pull request?

artichikin:
They merged our new one into master on Feb 10th.

PR #22, opened 2014-12-09
Zendesk.lookupOrganizationByExternalId no longers returns a PagedIterable

berttheuwissen:
...e because the Zendesk API doesn not support paging on the /api/v2/organizations/search.json?external_id={external_id} endpoint.


johnou:
It does support pagination but there was a bug in the PagedIterator, thanks for the report.

PR #27, opened 2015-01-05
Upgrade async-http-client to 1.9.3.

### LINKED ISSUE #26: Please update async-http-client to 1.9.x
Hello, we have another dependency that requires async-http-client 1.9.x.  It would be great if zendesk-java-client could update to 1.9.x, there are only a few breaking interface changes that impact zendesk-java-client. 
```xml
    <dependency>
      <groupId>com.ning</groupId>
      <artifactId>async-http-client</artifactId>
      <version>1.7.12</version>
    </dependency>
```
1.9.3 is the latest version.

Thanks!

abhinav:
Hey

This upgrades the version of async-http-client used by the library to 1.9.3.

Two breaking changes had to be fixed:

- `AsyncHttpClient.executeRequest` no longer throws `IOException`s. Instead they're just passed down to the handler. To fix this I added `ZendeskCompletionHandler` whose `onThrowable` wraps the `IOException`s in a `ZendeskException`.
- `RequestBuilder.addQueryParameter` was renamed to `addQueryParam`. Straightforward fix.

I wasn't sure if this merits a major version bump because of the backwards incompatible change in a dependency. I'll leave that decision to you.

This resolves #26.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
     <dependency>
       <groupId>com.ning</groupId>
       <artifactId>async-http-client</artifactId>
-      <version>1.7.12</version>
+      <version>1.9.3</version>
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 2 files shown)


stephenc:
Hmmm I'll need to evaluate this w.r.t. backwards compat

johnou:
@stephenc something like this? https://github.com/johnou/zendesk-java-client/commit/7a5f3d299f3844d7c78dc610a53cb8bccf787e2c

johnou:
Closing in favour of backwards compat solution, thanks for taking the time to create the pull @abhinav.

PR #28, opened 2015-01-21
Adding support for querying TicketForms.

gvh1234:
Hi,
here's a new feature I had to add to your code to support my company's integration work between our java app and Zendesk. We use a lot of custom forms in Zendesk and wanted to mirror them in our Java app. It didn't take much to get it working.

getTicketForms() and getTicketForm(id)

thanks for all the work


matthewhillier:
This is just what we're looking for. Thanks for this @gvh1234!

PR #35, opened 2015-02-23
Support for macros without restrictions.

johnou:
Please consider merging this before creating a new release to avoid breaking changes.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Action.java
+++ b/src/main/java/org/zendesk/client/v2/model/Action.java
@@ -1,0 +1,83 @@
+package org.zendesk.client.v2.model;
+
+import com.fasterxml.jackson.core.JsonParser;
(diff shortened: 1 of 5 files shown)


src/main/java/org/zendesk/client/v2/model/Action.java
```diff
@@ -0,0 +1,80 @@
...
+    public String toString() {
+        final StringBuilder sb = new StringBuilder();
+        sb.append("Action");
+        sb.append("{field=").append(field);
+        sb.append(", value=").append(value);
+        sb.append('}');
+        return sb.toString();
+    }
+
+    static class ActionDeserializer extends JsonDeserializer<Action> {
```
johnou: changed from public to default as there is no need to expose this inner class.

PR #37, opened 2015-02-24
Fix root cause of infinite looping "bug" in PagedIterable.

johnou:


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1463,11 +1444,8 @@ public boolean hasNext() {
                     if (page > 0) {                       
                         List<T> values = complete(submit(req("GET", url, page++), handler));
                         if (values == null || values.isEmpty()) {
-                            page = -1;
```
johnou: "bug" because it looks like the infinite looping was coded intentionally but doesn't seem desirable..

johnou:
Ignore this, will ask artichikin for more details. Current code works fine with my tests.

PR #39, opened 2015-02-25
Fix infinite paged iterable bug.

johnou:


pom.xml
```diff
@@ -20,12 +20,12 @@
   <parent>
     <groupId>com.cloudbees</groupId>
     <artifactId>cloudbees-oss-parent</artifactId>
-    <version>5</version>
+    <version>6</version>
```
johnou: You can ignore this, should be the same values but I had to merge master to my branch.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1103,42 +1102,34 @@ public void updateInstallation(int id, String json) {
...
-    private Request req(String method, Uri template, int page) {
+    private Request req(String method, Uri template, String contentType, byte[] body) {
         RequestBuilder builder = new RequestBuilder(method);
         if (realm != null) {
             builder.setRealm(realm);
         } else {
             builder.addHeader("Authorization", "Bearer " + oauthToken);
         }
-        builder.addQueryParameter("page", Integer.toString(page));
-        builder.setUrl(template.toString().replace("%2B", "+")); //replace out %2B with + due to API restriction
```
johnou: optimized API restriction replacement as well, now it doesn't compile a new pattern every request.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1431,48 +1416,34 @@ public static ObjectMapper createMapper() {
...
             this.url = url;
-            this.initialPage = initialPage;
         }
 
         public Iterator<T> iterator() {
-            return new PagedIterator(initialPage);
+            return new PagedIterator(url);
         }
 
         private class PagedIterator implements Iterator<T> {
```
johnou: Unfortunately PagedIterator is too tightly coupled to Zendesk and I didn't have time to refactor / extract it.

stephenc: I don't see that it should be decoupled from Zendesk necessarily

PR #41, opened 2015-02-26
Allow users to specify async-http-client version (1.9.3 compatible).

johnou:
@stephenc didn't want to merge this in without first discussing it with you, please let me know your thoughts.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,6 +27,10 @@
   <artifactId>zendesk-java-client</artifactId>
   <version>0.2.5-SNAPSHOT</version>
 
+  <properties>
+    <async-http-client.version>1.7.12</async-http-client.version>
+  </properties>
+
   <name>zendesk-java-client</name>
   <description>Java client for the Zendesk API</description>
   <url>https://github.com/cloudbees/zendesk-java-client</url>
@@ -71,7 +75,7 @@
     <dependency>
       <groupId>com.ning</groupId>
       <artifactId>async-http-client</artifactId>
-      <version>1.7.12</version>
+      <version>${async-http-client.version}</version>
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -1102,7 +1102,7 @@ public class Zendesk implements Closeable {
         }
     }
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1113,8 +1113,19 @@ public void updateInstallation(int id, String json) {
...
-            throw new ZendeskException(e.getMessage(), e);
+        } catch (Throwable t) {
+            throw new ZendeskException(t);
+        }
+    }
+
+    private static abstract class ZendeskAsyncCompletionHandler<T> extends AsyncCompletionHandler<T> {
+        @Override
+        public void onThrowable(Throwable t) {
+            if (t instanceof IOException) {
```
johnou: might be worth logging something here or throwing ZendeskException for all exceptions, super just chews it / logs debug.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1113,8 +1113,19 @@ public void updateInstallation(int id, String json) {
                 logger.debug("Request {} {}", request.getMethod(), request.getUrl());
             }
             return client.executeRequest(request, handler);
```
johnou: catching throwable here as 1.9.3 no longer throws IOException so we can support 1.7.12 (which anyway caught everything and wrapped it in IOException).

johnou: @stephenc what did you need to check wrt compatibility, I would be for removing the try catch and bumping major version.

stephenc: checked, let's go for it

pom.xml
```diff
@@ -71,7 +75,7 @@
     <dependency>
       <groupId>com.ning</groupId>
       <artifactId>async-http-client</artifactId>
-      <version>1.7.12</version>
+      <version>${async-http-client.version}</version>
```
stephenc: don't see the need for a property as we only have one dep.

pom.xml
```diff
@@ -27,6 +27,10 @@
   <artifactId>zendesk-java-client</artifactId>
   <version>0.2.5-SNAPSHOT</version>
 
+  <properties>
```
stephenc: properties are only for use when you reuse the dependency version in multiple places in the same pom or where you want to test multiple versions from the cli

PR #42, opened 2015-03-06
Incremental Export and User by Role support

Hillgod:
Added support for the incremental export API.
Also added support for querying for users by role.

Updated README.md to reflect the incremental export API.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -242,9 +242,17 @@ public class Zendesk implements Closeable {
         return new PagedIterable<Ticket>(cnst("/tickets/recent.json"), handleList(Ticket.class, "tickets"));
     }
     
-    public Iterable<Ticket> getTicketsIncrementally(Date fromDate) {
+    public Iterable<Ticket> getTicketsIncrementally(Date startTime) {
         return new PagedIterable<Ticket>(
-                tmpl("/incremental/tickets.json{?start_time}").set("start_time", (int) (fromDate.getTime() / 1000)), 
+                tmpl("/incremental/tickets.json{?start_time}").set("start_time", (int) (startTime.getTime() / 1000)), 
+                handleList(Ticket.class, "tickets"));                
+    }
+    
+    public Iterable<Ticket> getTicketsIncrementally(Date startTime, Date endTime) {
+        return new PagedIterable<Ticket>(
+                tmpl("/incremental/tickets.json{?start_time,end_time}")
+                    .set("start_time", (int) (startTime.getTime() / 1000))
+                    .set("end_time", (int) (endTime.getTime() / 1000)), 
                 handleList(Ticket.class, "tickets"));                
     }
 
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -6,7 +6,6 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
-import com.google.common.util.concurrent.RateLimiter;
 import com.ning.http.client.AsyncCompletionHandler;
 import com.ning.http.client.AsyncHttpClient;
 import com.ning.http.client.ListenableFuture;
@@ -58,6 +57,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
 /**
@@ -1241,7 +1241,7 @@ public class Zendesk implements Closeable {
 
     private static final String NEXT_PAGE = "next_page";
     private static final String END_TIME = "end_time";
-    private static final long FIVE_MINUTES_IN_MS = 300000;
+    private static final long FIVE_MINUTES_IN_MS = TimeUnit.MINUTES.toMillis(5);
 
     private abstract class PagedAsyncCompletionHandler<T> extends AsyncCompletionHandler<T> {
         String nextPage;
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -1245,7 +1245,7 @@ public class Zendesk implements Closeable {
     private static final long FIVE_MINUTES_IN_MS = TimeUnit.MINUTES.toMillis(5);
 
     private abstract class PagedAsyncCompletionHandler<T> extends AsyncCompletionHandler<T> {
-        String nextPage;
+        private String nextPage;
 
         public void setPagedProperties(JsonNode responseNode, Class<?> clazz) {
             JsonNode node = responseNode.get(NEXT_PAGE);
@@ -1259,11 +1259,15 @@ public class Zendesk implements Closeable {
         public String getNextPage() {
             return nextPage;
         }
+        
+        public void setNextPage(String nextPage) {
+            this.nextPage = nextPage;
+        }
     }
     
     private class PagedAsyncListCompletionHandler<T> extends PagedAsyncCompletionHandler<List<T>> {
-        final Class<T> clazz;
-        final String name;
+        private final Class<T> clazz;
+        private final String name;
         public PagedAsyncListCompletionHandler(Class<T> clazz, String name) {
             this.clazz = clazz;
             this.name = name;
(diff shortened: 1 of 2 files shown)


johnou:
Rate limiting is an interesting one, I'd like to hear @stephenc 's thoughts on it but I am happy to share our rate limiting solution.

```java
AsyncHttpClientConfig config = new AsyncHttpClientConfig.Builder().addRequestFilter(new RateLimiterRequestFilter()).build();
```

```java
/**
 * {@link com.ning.http.client.filter.RequestFilter} which is tuned to Zendesk's API rate limit (200 rpm).
 *
 * @author Johno Crawford (johno@sulake.com)
 */
public class SimpleRateLimiterRequestFilter implements RequestFilter {

    private RateLimiter rateLimiter = RateLimiter.create(3.0);

    @Override
    public FilterContext filter(FilterContext context) throws FilterException {
        rateLimiter.acquire();
        return new FilterContext.FilterContextBuilder(context).build();
    }
}
```

```java
/**
 * {@link com.ning.http.client.filter.RequestFilter} which is tuned to Zendesk's API rate limit (200 rpm / endpoint).
 *
 * @author Johno Crawford (johno@sulake.com)
 */
public class RateLimiterRequestFilter implements RequestFilter {

    private static final Logger logger = Logger.getLogger(RateLimiterRequestFilter.class);

    private static final ConcurrentHashMap<byte[], RateLimiter> rateLimiters = new ConcurrentHashMap<>();

    @Override
    public FilterContext filter(FilterContext context) throws FilterException {
        RateLimiter rateLimiter = getRateLimiter(context.getRequest().getInetAddress().getAddress());
        double secondsWaited = rateLimiter.acquire();
        if (secondsWaited > 0) {
            logger.info("Request throttled, waited " + secondsWaited + " seconds.");
        }
        return new FilterContext.FilterContextBuilder(context).build();
    }

    private RateLimiter getRateLimiter(byte[] remoteAddress) {
        RateLimiter rateLimiter = rateLimiters.get(remoteAddress);
        if (rateLimiter != null) {
            return rateLimiter;
        }
        rateLimiter = RateLimiter.create(3.0);
        RateLimiter oldRateLimiter = rateLimiters.putIfAbsent(remoteAddress, rateLimiter);
        return oldRateLimiter != null ? oldRateLimiter : rateLimiter;
    }
}
```

Hillgod:
Egh, I didn't intend for that to be part of this pull request. It's a fairly naive implementation of it. I'll see what I can do downstream to get it out of this pull request, at least.

stephenc:
@johnou your rate limiter looks good

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1195,9 +1240,11 @@ public T onCompleted(Response response) throws Exception {
     }
 
     private static final String NEXT_PAGE = "next_page";
+    private static final String END_TIME = "end_time";
+    private static final long FIVE_MINUTES_IN_MS = 300000;
```
johnou: Might be a little nicer to read using TimeUnit eg. TimeUnit.MINUTES.toMillis(5)

Hillgod: Cool, made that change.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -434,6 +462,23 @@ public void deleteTrigger(long triggerId) {
...
         return new PagedIterable<User>(cnst("/users.json"), handleList(User.class, "users"));
     }
+    
+    public Iterable<User> getUsersByRole(String role, String... roles) {
+        // Going to have to build this URI manually, because the RFC6570 template spec doesn't support
+        // variables like ?role[]=...role[]=..., which is what Zendesk requires.
+        // See https://developer.zendesk.com/rest_api/docs/core/users#filters
+        final StringBuilder uriBuilder = new StringBuilder("/users.json");
+        if (roles.length == 0) {
+            uriBuilder.append("?role=").append(role);
```
johnou: do we need to uri encode the role?

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -228,6 +242,20 @@ public void deleteTickets(long id, long... ids) {
...
+    public Iterable<Ticket> getTicketsIncrementally(Date startTime) {
+        return new PagedIterable<Ticket>(
+                tmpl("/incremental/tickets.json{?start_time}").set("start_time", (int) (startTime.getTime() / 1000)), 
+                handleIncrementalList(Ticket.class, "tickets"));                
+    }
+    
+    public Iterable<Ticket> getTicketsIncrementally(Date startTime, Date endTime) {
+        return new PagedIterable<Ticket>(
+                tmpl("/incremental/tickets.json{?start_time,end_time}")
+                    .set("start_time", (int) (startTime.getTime() / 1000))
```
johnou: for readability I'd suggest creating a small private method which maybe takes a date divides by 1000 and casts, looks a ugly over and over.

johnou:
When I get some time I'll look at trying to integrate something like http://blog.nirav.name/2015/02/a-simple-rate-limiter-using-javas.html into the project instead of adding a dependency on guava..

johnou:
@Hillgod would you have time to implement the remaining endpoints listed https://developer.zendesk.com/rest_api/docs/core/incremental_export if not please mark which ones are not implemented in the README.

stephenc:
@johnou yes, let's not add guava as a dependency... that project has a completely different idea of backwards compatibility from everyone else... depending on it in a utility jar (such as this) is a really bad plan... such a pity that we have to run NIH syndrome with guava as it would otherwise be a lovely library... sadly I have been burned 5 times too often with guava

Hillgod:
I'll make the suggested code changes ASAP. I'm about to be on vacation for over a week, so the other incremental exports will have to wait until at least then, but I'm probably up for it. Regardless, I'll update the README.md to better reflect the new situation w.r.t that API.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -434,6 +462,23 @@ public void deleteTrigger(long triggerId) {
...
+        final StringBuilder uriBuilder = new StringBuilder("/users.json");
+        if (roles.length == 0) {
+            uriBuilder.append("?role=").append(role);
+        } else {
+            uriBuilder.append("?role[]=").append(role);
+        }
+        for (final String curRole : roles) {
+            uriBuilder.append("&role[]=").append(curRole);
+        }
+        logger.info(uriBuilder.toString());
```
johnou: remove debug logging?

Hillgod:
I think I've addressed all concerns. Please let me know if you'd like to be do anything else before merging this in. Thanks!

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1212,23 +1260,53 @@ public String getNextPage() {
...
+                if (node == null) {
+                    throw new NullPointerException(NEXT_PAGE + " property not found, pagination not supported" +
+                            (clazz != null ? " for " + clazz.getName() : ""));
+                }
+                JsonNode endTimeNode = responseNode.get(END_TIME);
+                if (endTimeNode == null || endTimeNode.asLong() == 0) {
+                    throw new NullPointerException(END_TIME + " property not found, incremental export pagination not supported" +
+                            (clazz != null ? " for " + clazz.getName() : ""));
+                }
+                if (endTimeNode.asLong() * 1000 > System.currentTimeMillis() - FIVE_MINUTES_IN_MS) {
```
stephenc: You could use `TimeUnit.SECONDS.toMillis(endTimeNode.asLong())` rather than the less expressive `* 1000`

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1294,6 +1372,20 @@ private void logResponse(Response response) throws IOException {
...
+        try {
+            return URLEncoder.encode(str, UTF_8);
+        } catch (UnsupportedEncodingException e) {
+            // This really can't happen. UTF-8 is guaranteed to exist.
+            return str;
+        }
+    }
+
+    private static int msToSeconds(long millis) {
+        return (int) (millis / 1000);
```
stephenc: I'd rather `TimeUnit.MILLISECONDS.toSeconds()` which can safely be in-lined

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1528,8 +1620,8 @@ public Builder setOauthToken(String oauthToken) {
             }
             return this;
         }
-
-
+
```
stephenc: unnecessary whitespace changes

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1195,9 +1241,11 @@ public T onCompleted(Response response) throws Exception {
     }
 
     private static final String NEXT_PAGE = "next_page";
+    private static final String END_TIME = "end_time";
+    private static final long FIVE_MINUTES_IN_MS = TimeUnit.MINUTES.toMillis(5);
 
     private abstract class PagedAsyncCompletionHandler<T> extends AsyncCompletionHandler<T> {
-        private String nextPage;
+        String nextPage;
```
stephenc: does this really need escalation to `/*package*/` scope? if so, I'd prefer a getter/setter and keep it `private`

Hillgod: The reason I just made to `package` scope was we're already dealing with a private class, and an anonymous class below needs to set the value. But I've made it private and added a setter.

johnou:
Merged, thanks @Hillgod.

PR #62, opened 2015-07-29
Add api to search for articles in help center.

christ66:
Two changes were made to make searching for an article in the help center.

1. Make `Article` a `SearchResultEntity`, and add it to the list of available searchables.
2. Invoking the API call to the help center search.

@reviewbybees


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -92,6 +92,7 @@
        result.put("group", Group.class);
        result.put("organization", Organization.class);
        result.put("topic", Topic.class);
+        result.put("article", Article.class);
```
oleg-nenashev: NIT: indenting

oleg-nenashev:
:bee: as far as I understand it

PR #64, opened 2015-08-13
Update Zendesk.java

mamsterla:
Fixed the template for URI to correctly accept passed in token

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -428,7 +428,7 @@ public class Zendesk implements Closeable {
     }
 
     public Attachment.Upload createUpload(String token, String fileName, String contentType, byte[] content) {
-        TemplateUri uri = tmpl("/uploads.json{?filename}{?token}").set("filename", fileName);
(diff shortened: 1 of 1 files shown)


mamsterla:
This simple change should allow repeated uploads using the same token

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -428,7 +428,7 @@ public void deleteTicketField(long id) {
     }
 
     public Attachment.Upload createUpload(String token, String fileName, String contentType, byte[] content) {
-        TemplateUri uri = tmpl("/uploads.json{?filename}{?token}").set("filename", fileName);
+        TemplateUri uri = tmpl("/uploads.json{?filename}{&token}").set("filename", fileName);
```
stephenc: probably should be `{?filename,token}` instead

ydubreuil: I confirm, see here: https://github.com/cloudbees/zendesk-java-client/blob/master/src/main/java/org/zendesk/client/v2/Zendesk.java#L308

stephenc:
I have made the changes in #123. Thank you for your contribution.

PR #65, opened 2015-08-20
Support for bulk updating of tickets

jon-ruckwood:
I've added support for both variants of [bulk updating tickets][1]. See:

* `updateTickets(List<Ticket> tickets)` 
* `updateTickets(Ticket ticket, Collection<Long> ids)`

There are also additional variants to match the existing pattern for asynchronous API endpoints.

[1]: https://developer.zendesk.com/rest_api/docs/core/tickets#bulk-updating-tickets


jon-ruckwood:
For reference this was implemented in PR #260

PR #48, opened 2015-04-17
Adding implements Serializable to the model beans

gvh1234:
To use with Spring Redis caching

PR #66, opened 2015-08-20
Implement Serializable for Model classes

matthewtckr:
Submitting an updated PR for #48


matthewtckr:
Rebased and set a `serialVersionUID` for each class now implementing the Serializable interface.

PR #74, opened 2015-10-23
Apply a consistent code style so that PRs and code reading is easier

peterlynch:
The code style applied is based on templates at https://github.com/sonatype/codestyle

I don't want to debate semantics of style but at least should have something and the above is publicly available and consumable by popular IDEs.


peterlynch:
@stephenc do you have a code style you would prefer to apply or should I just close this

stephenc:
:-1: on this style

stephenc:
https://github.com/cloudbees/zendesk-java-client/blob/master/CONTRIBUTING.adoc

PR #76, opened 2015-10-25
Changed userId to Long to accommodate values larger than Integer.MAX_…

mttdbrd:
…INT.

As of October 24, 2015, Zendesk userIds are already larger than Integer.MAX_INT. This means that for any new client using Zendesk Java code, user IDs in the Identity class will be negative numbers. Most other classes use Long for ids, but this one does not.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Identity.java
+++ b/src/main/java/org/zendesk/client/v2/model/Identity.java
@@ -11,7 +11,7 @@ import java.util.Date;
 public class Identity {
     private Long id;
     private String url;
-    private Integer userId;
+    private Long userId;
     private String type;
     private String value;
     private Boolean verified;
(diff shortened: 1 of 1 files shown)


christ66:
@mttdbrd Has it been reported to zendesk that their API is still listing integer for the user ID (https://developer.zendesk.com/rest_api/docs/core/user_identities)?

mttdbrd:
@christ66 The id field for User (https://developer.zendesk.com/rest_api/docs/core/users) also says integer, but the zendesk-java-client uses Java long. I don't think their documentation is incorrect to use integer. Integer has a range of meanings in different programming languages.

stephenc:
need to also bump the minor version number as this is a binary incompatible change (it would be a major bump but we are still in `0.x.y` land

aheritier:
@stephenc I know that for another tool we have internally (zendesk bot) which isn't yet using this library I had to do such change because various identifiers are already > Integer.MAX_INT
Maybe we should release a last minor version with changes that doesn't change the model and then bump to a new major to release with this change

PR #82, opened 2015-11-10
Incremental Articles

matthewtckr:


matthewtckr:
This is missing a smoke test, as passing a `new Date(0L)` to the API may cause HTTP 502 errors for certain Zendesk accounts.  In speaking with Zendesk Support, it sounds like the minimum value allowed is dependent upon when the account started using the HelpCenter.  This is different than the Incremental Tickets API, which allows for a `0` start_time.


```
public static void main( String[] args ) {
    Zendesk client = new Zendesk.Builder("https://{subdomain}.zendesk.com/").setUsername( "XXX" ).setToken( "XXX" ).build();

    //Iterable<Article> articles = client.getArticlesIncrementally( new Date( 1444219200000L ) );
    Iterable<Article> articles = client.getArticlesIncrementally( new Date( 0L ) );
    int i = 0;
    for (Article article : articles ) {
      i++;
      System.out.println( article.getUpdatedAt().toString() + ": " + article.getTitle() );
    }
    System.out.println( "Total: " + i );
}
```

matthewtckr:
The Incremental Articles API returns null values for next_page and end_time, which causes an NPE to be thrown in cases where no articles are being returned.  For instance, if the API is called with `new Date()`, it's not likely that an article was created at that exact time.

Added smoke test, including handling for the API limitation mentioned above.

ydubreuil:
@matthewtckr could you rebase on current master?

matthewtckr:
@ydubreuil Rebased.  Thanks for all of the recent commits, hope you guys had a great holiday!

PR #83, opened 2015-11-13
Code Style Only

peterlynch:
Took the default IntelliJ Code style, applied rules outlined in CONTRIBUTING.adoc, then applied code formatting to all Java src files in src/main.


ydubreuil:
Changing code formatting makes it difficult to understand code history. Closing this PR.

peterlynch:
What is the point of having a code style if you never enforce it?

While I agree with your statement in regards historical changes, long term understanding the code becomes easier and makes PRs less of a problem when a consistent style is eventually applied one time.

PR #89, opened 2015-12-15
createOAuthToken and updateTarget endpoints

patrikvarga:
I added the createOAuthToken and updateTarget endpoints and fix some issues.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -45,6 +45,8 @@ import org.zendesk.client.v2.model.hc.Article;
 import org.zendesk.client.v2.model.hc.ArticleAttachments;
 import org.zendesk.client.v2.model.hc.Category;
 import org.zendesk.client.v2.model.hc.Section;
+import org.zendesk.client.v2.model.oauth.OAuthTokenRequest;
+import org.zendesk.client.v2.model.oauth.OAuthTokenResponse;
 import org.zendesk.client.v2.model.targets.BasecampTarget;
 import org.zendesk.client.v2.model.targets.CampfireTarget;
 import org.zendesk.client.v2.model.targets.EmailTarget;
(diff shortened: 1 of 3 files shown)


ydubreuil:
Thanks for your PR. Could you revert autoformat commit so that only changes related to the feature remain?

patrikvarga:
I was just following the Code Style Guidelines described in CONTRIBUTING.adoc. ;)
The cloudbees:master code does not comply with its own formatting guidelines, my commit does (more).

Example violations:
* "To the largest extent possible, all fields shall be private." (This is relevant to another commit of mine: fe02e50.)
* "Java blocks are 4 spaces." (This is the commit in question: 319ecd5.)

That said, those commits (and 48a2819) indeed are not relevant to the PR, so if you still think I should revert, I can try get back to it in a few days. Thanks.

ydubreuil:
> I was just following the Code Style Guidelines described in CONTRIBUTING.adoc. ;)

No worries. Generally speaking, it's better to avoid changing code style of existing code. It makes it difficult to browse code history for almost no benefits. Let's use Code Style Guidelines for new code only.

> That said, those commits (and 48a2819) indeed are not relevant to the PR, so if you still think I should revert, I can try get back to it in a few days. Thanks.

I would prefer that you revert them.

patrikvarga:
OK, I'll get back to it soon. (I noticed those commits after creating the PR. Agreed that they have no place in this PR.)

> Generally speaking, it's better to avoid changing code style of existing code. It makes it difficult to browse code history for almost no benefits. Let's use Code Style Guidelines for new code only.

I strongly disagree with this. Using the Code Style Guidelines for new code only makes no sense in this codebase which is essentially one large file. Making changes is a pain as everyone is forced to disable auto-format, format-on-save etc. (even though they *are* compliant with the guidelines), thus effectively rendering the guidelines useless.

patrikvarga:
Please see new PR #107

patrikvarga:
Re: "Fix search result type mapping bug" see PR #108

patrikvarga:
Closing this as there are other PRs mentioned above.

Please note that the formatting issue is still a burden to contributors and should be fixed.

PR #92, opened 2015-12-30
HelpCenter Add paging for large resultset

matthewtckr:


ydubreuil:
I'm worried it will break backward compatibility.

We could merge it in 0.4 with other breaking changes. WDYT @stephenc @aheritier ? I'm gonna create a 0.4 milestone to track these breaking changes if that works for you.

stephenc:
@ydubreuil I think we should be ready to bump to 0.4 now... if you think a release is needed first we can discuss

ydubreuil:
I've merged few non breaking PR, so we can release a last 0.3 version, but I'm not sure it really makes sense. I can bump master to 0.4 and merges remaining PR

ydubreuil:
@matthewtckr merged all PR targeting 0.4, will need to do some tests before releasing

PR #116, opened 2016-06-13
Fix getting search results' classes

boris-petrov:
Without this, the getSearchResults(String) method does not work at all.


patrikvarga:
Same as PR #108

stephenc:
#108 was merged. Thank you for your contribution

PR #117, opened 2016-06-15
Add Get / Create methods for Help Center translations 

dlyash:


stephenc:
I'll need to run this on the test account Zendesk gave us to verify that the tests pass. Otherwise looks good

Issue #126, opened 2016-08-04
New AgentMacroReference Event Type


matthewtckr:
Came across this in our ETL process, it appears to be a new Zendesk Event type that's not implemented, and apparently no documentation from Zendesk.

Sample Event from a ticket audit entry:

```
{
            "id": 121204214906,
            "type": "AgentMacroReference",
            "via": {
                "channel": "web",
                "source": {
                    "from": {},
                    "to": {},
                    "rel": null
                }
            },
            "macro_title": "Workflow Stage: Problem Definition",
            "macro_id": "49985043"
        }
```

Sample error for unrecognized event (Zendesk Java Client 0.4.0):

```
Unexpected error
org.zendesk.client.v2.ZendeskException: java.lang.IllegalArgumentException: Could not resolve type id 'AgentMacroReference' into a subtype of [simple type, class org.zendesk.client.v2.model.events.Event]
 at [Source: N/A; line: -1, column: -1] (through reference chain: org.zendesk.client.v2.model.Audit["events"])
    at org.zendesk.client.v2.Zendesk.complete(Zendesk.java:1873)
    at org.zendesk.client.v2.Zendesk.access$1300(Zendesk.java:83)
    at org.zendesk.client.v2.Zendesk$PagedIterable$PagedIterator.hasNext(Zendesk.java:2055)
    at org.pentaho.di.trans.steps.zendesk.ZendeskInputTicketAudit.processRow(ZendeskInputTicketAudit.java:115)
    at org.pentaho.di.trans.step.RunThread.run(RunThread.java:62)
    at java.lang.Thread.run(Thread.java:745)
Caused by: java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: Could not resolve type id 'AgentMacroReference' into a subtype of [simple type, class org.zendesk.client.v2.model.events.Event]
 at [Source: N/A; line: -1, column: -1] (through reference chain: org.zendesk.client.v2.model.Audit["events"])
    at com.ning.http.client.providers.netty.future.NettyResponseFuture.done(NettyResponseFuture.java:220)
    at com.ning.http.client.providers.netty.handler.HttpProtocol.finishUpdate(HttpProtocol.java:194)
    at com.ning.http.client.providers.netty.handler.HttpProtocol.handleChunk(HttpProtocol.java:451)
    at com.ning.http.client.providers.netty.handler.HttpProtocol.handle(HttpProtocol.java:474)
[...]

PR #127, opened 2016-08-05
AgentMacroReference Event Type

matthewtckr:
Issue #126

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/events/AgentMacroReferenceEvent.java
+++ b/src/main/java/org/zendesk/client/v2/model/events/AgentMacroReferenceEvent.java
@@ -1,0 +1,46 @@
+package org.zendesk.client.v2.model.events;
+
+import org.zendesk.client.v2.model.Via;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class AgentMacroReferenceEvent extends Event {
+
+    private static final long serialVersionUID = 1L;
+
+    private Long macroId;
+    private String macroTitle;
+    private Via via;
+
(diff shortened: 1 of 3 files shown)


src/main/java/org/zendesk/client/v2/model/events/AgentMacroReferenceEvent.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.zendesk.client.v2.model.Via;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class AgentMacroReferenceEvent extends Event {
+
+    private static final long serialVersionUID = 1L;
+
+    private Long macroId;
```
matthewtckr: I'm hesitant to set this as a String (which is what is used in the JSON), I'll submit a few Zendesk Support cases tomorrow for some documentation from Zendesk.

jchambers:
:metal:

This just started making life difficult for us, too.

jchambers:
One other thought: would it make sense to have a `defaultImpl` in the `JsonTypeInfo` annotation in `Event.java`? That way, we can fall back to a not-exactly-desirable-but-at-least-not-catastrophic `UnknownEvent` if new event types get introduced upstream.

matthewtckr:
@jchambers Submitted PR #128 to handle these new, unknown event types.  Feel free to take a look and provide any feedback there.

I agree that `<sarcasm>it's extremely satisfying</sarcasm>` to get notifications like this after-hours because of a Zendesk change.

rodrigopalhares:
Nice job!!! Thx!

PR #144, opened 2016-08-25
Add separate exception type for HTTP 404

dlyash:


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1652,7 +1652,7 @@ public T onCompleted(Response response) throws Exception {
                 if (response.getStatusCode() == 404) {
                     return null;
```
ydubreuil: Convention is to return `null` when there's a 404, like here. The new exception will be not thrown here.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1750,7 +1750,7 @@ public PagedAsyncListCompletionHandler(Class<T> clazz, String name) {
                 }
                 return values;
             }
-            throw new ZendeskResponseException(response);
+            throw ZendeskResponseException.fromResponse(response);
```
ydubreuil: Instead of throwing the exception for a list, I think it could be better to return an empty list, and treat this case as a normal case.

PR #145, opened 2016-08-25
Fix #143

### LINKED ISSUE #143: Adding support for create or update user
https://developer.zendesk.com/rest_api/docs/core/users#create-or-update-user

KevinIp:
- Add support for create or update user

### DIFF
--- a/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
+++ b/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
@@ -671,4 +671,44 @@ public class RealSmokeTest {
             }
         }
     }
+
+    @Test
+    public void createOrUpdateUser() throws Exception {
+        createClientWithTokenOrPassword();
+
+        String name = "testCreateOrUpdateUser";
+        String externalId = "testCreateOrUpdateUser";
+
+        // Clean up to avoid conflicts
+        for (User u: instance.lookupUserByExternalId(externalId)){
+            instance.deleteUser(u.getId());
+        }
+
+        String phoneAtCreation = "5555551234";
+        User user = new User(true, name);
+        user.setExternalId(externalId);
+        user.setPhone(phoneAtCreation);
+
+        User createResult = instance.createOrUpdateUser(user);
+        assertNotNull(createResult);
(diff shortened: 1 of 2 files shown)


KevinIp:
@stephenc, could you help take a look when you get a chance? Thanks

src/test/java/org/zendesk/client/v2/RealSmokeTest.java
```diff
@@ -671,4 +671,45 @@ public void getSections() throws Exception {
...
     }
+
+    @Test
+    public void createOrUpdateUser() throws Exception {
+        createClientWithTokenOrPassword();
+
+        String name = "testCreateOrUpdateUser";
+        String externalId = "testCreateOrUpdateUser";
+        // Clean up to avoid conflicts
+        for (User u : instance.getUsers()) {
```
johnou: Instead of iterating all users could we not just look it up by external id?

johnou:
Might be nice to implement all options for this endpoint to avoid possible API breakage and useless overloads. 

https://developer.zendesk.com/rest_api/docs/core/users#create-or-update-user

"If you need to create users without sending out a verification email, pass a "verified": true parameter."

KevinIp:
Good call. It seems the parameter is for the json object, while having it as an URL parameter has no effect.

```
curl -v -u "$credential" "$url/users/create_or_update.json" \
  -H "Content-Type: application/json" -X POST -d '{"user": {"name": "test", "verified": true, "email": "test1@example.com"}}' | jq

{
  "user": {
    "id": 12573044948,
    "verified": true,
    ... //omit for clarity
  }
}
```

```
curl -v -u "$credential" "$url/users/create_or_update.json?verified=true" \
  -H "Content-Type: application/json" -X POST -d '{"user": {"name": "test", "email": "test20@example.com"}}' | jq

{
  "user": {
    "id": 12573301608,
    "verified": false,
    ...
  }
}
```

src/test/java/org/zendesk/client/v2/RealSmokeTest.java
KevinIp: updated

ydubreuil:
LGTM

PR #154, opened 2016-10-27
Add get articles by category

rifqifatih:
[Reference](https://developer.zendesk.com/rest_api/docs/help_center/articles)

I could write the unit tests as well, but there seems to be no convention yet for testing methods with parameter.


aheritier:
should be ok

ydubreuil:
LGTM

PR #158, opened 2016-11-09
Add brand to model. Support list brands api request.

satamas:


aheritier:
It seems ok to be merged @ydubreuil @stephenc @christ66

PR #164, opened 2017-02-02
Adding html and plain text fields to Comment.

erikrj:
As per the API documentation located at https://developer.zendesk.com/rest_api/docs/core/ticket_comments, there are html_body and plain_body are options for a Comment. I've added these two fields as we needed to be able to submit HTML formatted tickets. It would be nice if this could be merged so we don't have to depend on our forked version.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Comment.java
+++ b/src/main/java/org/zendesk/client/v2/model/Comment.java
@@ -17,6 +17,8 @@ public class Comment implements Serializable {
 
     private Long id;
     private String body;
+    private String plainBody;
+    private String htmlBody;
     private Long authorId;
     private List<String> uploads;
     private List<Attachment> attachments;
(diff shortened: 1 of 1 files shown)


kamilszymanski:
having access to HTML body would be really useful, waiting for this PR to be merged

cbrunnkvist:
works for me 👍

gvh1234:
I'd love this to be merged,
thanks

src/main/java/org/zendesk/client/v2/model/Comment.java
```diff
@@ -43,6 +45,24 @@ public void setBody(String body) {
         this.body = body;
     }
 
+    @JsonProperty("plain_body")
+    public String getPlainBody() {
+        return plainBody;
+    }
+
+    public void setPlainBody(String plainBody) {
```
johnou: API states plain_body is read only, setter might cause errors / confusion.

johnou:
Closed in favour of https://github.com/cloudbees/zendesk-java-client/pull/184 because https://github.com/cloudbees/zendesk-java-client/pull/164#discussion_r127032874 was not addressed.

PR #166, opened 2017-02-08
Add restrictedAgent property to User model.

satamas:


johnou:
https://developer.zendesk.com/rest_api/docs/core/users

johnou:
@satamas thanks!

PR #171, opened 2017-03-22
Add method to inline an article attachment.

christ66:
According to: https://developer.zendesk.com/rest_api/docs/help_center/article_attachments#create-article-attachment

the curl command:

```
curl https://{subdomain}.zendesk.com/api/v2/help_center/articles/{id}/attachments.json \
  -F "inline=true" -F "file=@drawing.png" \
  -v -u {email_address}:{password} -X POST
```

Lets you set the inline option when creating the attachment for the first time. It will allow you to create an attachment without it having to be part of the attachments page.

@reviewbybees

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/hc/ArticleAttachments.java
+++ b/src/main/java/org/zendesk/client/v2/model/hc/ArticleAttachments.java
@@ -9,7 +9,7 @@ public class ArticleAttachments {
     /**
      *  Automatically assigned when the article attachment is created
      */
-    private Integer id;
+    private Long id;
 
     /**
      *  The API url of this article attachment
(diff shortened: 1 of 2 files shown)


src/main/java/org/zendesk/client/v2/model/hc/ArticleAttachments.java
```diff
@@ -9,7 +9,7 @@
     /**
      *  Automatically assigned when the article attachment is created
      */
-    private int id;
+    private Integer id;
```
aheritier: https://developer.zendesk.com/rest_api/docs/help_center/article_attachments
All fields marked as integer in the doc should be Long in Java
See PR#170

johnou: @christ66 ping

PR #174, opened 2017-04-27
Field's Option constructor class made public.

russellhoff:
I've made Option's constructor public.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Field.java
+++ b/src/main/java/org/zendesk/client/v2/model/Field.java
@@ -265,10 +265,10 @@ public class Field implements Serializable {
         private String name;
         private String value;
 
-        private Option() {
+        public Option() {
         }
 
-        private Option(String name, String value) {
+        public Option(String name, String value) {
             this.name = name;
             this.value = value;
         }


russellhoff:
please, could you accept this pull request?

russellhoff:
What happened to the pull request?

russellhoff:
@stephenc something happened to jenkins...

russellhoff:
Hello @stephenc , could you tell me how to improve the code to be acceptable by jenkins?

stephenc:
@aheritier do you know where that jenkins is. I cannot access even when on vpn

aheritier:
strange @stephenc I have no idea
no idea @Dohbedoh ?
The one I use is https://support.ci.cloudbees.com/job/DEV/job/zendesk-java-client/

russellhoff:
Hello? Anything to say about that PR?

russellhoff:
But Jenkins said "All checks have failed"

russellhoff:
I'm waiting for any response....

russellhoff:
Hi there! What happened here? Have all of you been abducted by an alien? :p

gvh1234:
I believe this can be merged? It's causing pain

src/main/java/org/zendesk/client/v2/model/Field.java
```diff
@@ -265,10 +265,10 @@ public String toString() {
         private String name;
         private String value;
 
-        private Option() {
+        public Option() {
```
johnou: Why would you need to create an option with uninitialised values? This should only be required for Jackson and should be ok to leave as private.

russellhoff: Is there other way to construct Options? You tell me, please. I was developing a project which needed the construction of Options.

johnou: Yes, with the second constructor which takes the name and value args.

PR #175, opened 2017-05-22
Suspend User

matthewtckr:


matthewtckr:
This is added mainly as a convenience function compared to the updateUser method.  Also to ensure that only the `suspended` attribute is modified for a user when the method is called.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -722,6 +722,19 @@ public void deleteUser(long id) {
         complete(submit(req("DELETE", tmpl("/users/{id}.json").set("id", id)), handleStatus()));
     }
 
+    public User suspendUser(User user, boolean action) {
```
johnou: Let's keep it simple, one endpoint with user id.

johnou: suspendUser and unsuspendUser? Tough one, method named suspend user to unsuspend doesn't seem logical.

matthewtckr: I'm fine with either solution, let me know if you want me to change it (looks like you might still be undecided).

Try reading the Zendesk doc, they don't ever use the word "unsuspend", or "reinstate" or other antonym for "suspend". Tough decisions in trying to make a simple thing simple. :)

johnou: https://support.zendesk.com/hc/en-us/articles/203690976-Suspending-a-user

"To unsuspend a user, follow the steps above and select Unsuspend access."

johnou: @matthewtckr let's go with suspendUser and unsuspendUser with simple user id argument.

matthewtckr: @johnou Updated PR

PR #180, opened 2017-06-12
added created article,section,category translation

assafko:
added create translation for article,section,category
according to https://developer.zendesk.com/rest_api/docs/help_center/translations#create-translation

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -1551,7 +1551,7 @@ public class Zendesk implements Closeable {
     }
 
     public Translation createArticleTranslation(Long articleId, Translation translation) {
-        return complete(submit(req("POST", tmpl("/help_center/articles/{id}/translations/translations.json"),
+        return complete(submit(req("POST", tmpl("/help_center/articles/{id}/translations/translations.json").set("id", articleId),
                 JSON, json(Collections.singletonMap("translation", translation))), handle(Translation.class, "translation")));
     }
 
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -1551,6 +1551,7 @@ public class Zendesk implements Closeable {
     }
 
     public Translation createArticleTranslation(Long articleId, Translation translation) {
+        checkHasArticleId(articleId);
         return complete(submit(req("POST", tmpl("/help_center/articles/{id}/translations/translations.json").set("id", articleId),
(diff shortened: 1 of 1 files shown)


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1550,6 +1550,12 @@ public Article updateArticle(Article article) {
                 JSON, json(Collections.singletonMap("article", article))), handle(Article.class, "article")));
     }
 
+    public Translation createArticleTranslation(Long articleId, Translation translation) {
+        checkHasArticleId(articleId);
```
aheritier: Why do we validate the Id of the article while in updateArticleTranslation we are verifying the Id of translation ?

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1607,6 +1613,12 @@ public Category updateCategory(Category category) {
                 JSON, json(Collections.singletonMap("category", category))), handle(Category.class, "category")));
     }
 
+    public Translation createCategoryTranslation(Long categoryId, Translation translation) {
+        checkHasCategoryId(categoryId);
```
aheritier: Why do we validate the Id of the category while in updateCategoryTranslation we are verifying the Id of translation ?

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1652,6 +1664,12 @@ public Section updateSection(Section section) {
                 JSON, json(Collections.singletonMap("section", section))), handle(Section.class, "section")));
     }
 
+    public Translation createSectionTranslation(Long sectionId, Translation translation) {
+        checkHasSectionId(sectionId);
```
aheritier: Why do we validate the Id of the section while in updateSectionTranslation we are verifying the Id of translation ?

src/main/java/org/zendesk/client/v2/Zendesk.java
assafko: because in create the translation has no id (only post id)
but. we do need to make sure article id is not null

src/main/java/org/zendesk/client/v2/Zendesk.java
assafko: because in create article the translation has no id (only post id)
but. we do need to make sure article id is not null

src/main/java/org/zendesk/client/v2/Zendesk.java
assafko: because in create the translation has no id (only post id)
but. we do need to make sure section id is not null

PR #181, opened 2017-07-06
Change Field.Option to implement Serializable

gvh1234:
We need this because we cache data coming from Zendesk.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Field.java
+++ b/src/main/java/org/zendesk/client/v2/model/Field.java
@@ -262,6 +262,7 @@ public class Field implements Serializable {
     }
 
     public static class Option implements Serializable {
+        private static final long serialVersionUID = -8881532430230657120L;    
         private String name;
         private String value;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/zendesk/client/v2/model/Field.java
```diff
@@ -261,7 +261,7 @@ public String toString() {
         return sb.toString();
     }
 
-    public static class Option {
+    public static class Option implements Serializable {
```
johnou: Define serialVersionUID here or cached data may not be compatible between different versions of zendesk-java-client.

gvh1234: doh!

gvh1234: done

johnou:
Thanks for the contribution!

chuck-wood:
Woohoo, thanks for fixing this! One more thing I can cache :)

PR #183, opened 2017-07-17
Changed type of ticket field ID from Integer to Long.

ngyewch:
Fix for an ID value overflow problem.


johnou:
@ngyewch thanks!

ngyewch:
@johnou Thanks!

PR #185, opened 2017-07-26
Add additional attributes for Photo found in API responses

drd:
(Note: I use the models in your package separately from the client, because I am on Google App Engine, where only `URLFetch` is allowed. I'm unsure if the same issue would happen with your client!)

I started to get a bunch of errors like this while parsing user search results:

```
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "url" (class org.zendesk.client.v2.model.Attachment), not marked as ignorable (6 known properties: "content_type", "size", "thumbnails", "content_url", "id", "file_name"])
 at [Source: [B@70410355; line: 1, column: 316] (through reference chain: premise.proxy.mobile.dto.ZdUserSearchResult["users"]->java.util.ArrayList[0]->org.zendesk.client.v2.model.User["photo"]->org.zendesk.client.v2.model.Attachment["url"])
```

So, I looked at the raw JSON returned, and the shape is like this:

```
      "photo": {
        "url": "https://...",
        "id": 1040994207,
        "file_name": "...",
        "content_url": "https://..",
        "mapped_content_url": "https://....",
        "content_type": "image/png",
        "size": 12557,
        "width": 80,
        "height": 80,
        "inline": false,
        "thumbnails": [
          {
            "url": "https://....",
            "id": 1040994227,
            "file_name": "...",
            "content_url": "https://...",
            "mapped_content_url": "https://////",
            "content_type": "image/png",
            "size": 12557,
            "width": 32,
            "height": 32,
            "inline": false
          }
        ]
      },
```

So, I added fields for all of those missing (`url`, `mapped_content_url`, `width`, `height`, and `inline`), and to future-proof, I also added a `@JsonIgnoreProperties(ignoreUnknown = true)` annotation on the class.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Photo.java
+++ b/src/main/java/org/zendesk/client/v2/model/Photo.java
@@ -2,21 +2,28 @@ package org.zendesk.client.v2.model;
 
 import java.io.Serializable;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 /**
  * @author stephenc
  * @since 05/04/2013 15:36
  */
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class Photo implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
     private Long id;
     private String fileName;
+    private String url;
     private String contentUrl;
+    private String mappedContentUrl;
     private String contentType;
     private Long size;
+    private Long width;
+    private Long height;
+    private Boolean inline;
 
     @JsonProperty("content_type")
     public String getContentType() {
@@ -27,6 +34,14 @@ public class Photo implements Serializable {
         this.contentType = contentType;
     }
 
+    public String getUrl() {
+        return url;
+    }
+
+    public void setUrl(String url) {
+        this.url = url;
+    }
+
     @JsonProperty("content_url")
     public String getContentUrl() {
         return contentUrl;
@@ -34,6 +49,15 @@ public class Photo implements Serializable {
 
     public void setContentUrl(String contentUrl) {
         this.contentUrl = contentUrl;
+    }
+
+    @JsonProperty("mapped_content_url")
+    public String getMappedContentUrl() {
+        return mappedContentUrl;
+    }
+
+    public void setMappedContentUrl(String mappedContentUrl) {
+        this.mappedContentUrl = mappedContentUrl;
     }
 
     @JsonProperty("file_name")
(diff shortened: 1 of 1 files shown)


gvh1234:
This is more of a observation than a problem:
if we start down the path of using @JsonIgnoreProperties(ignoreUnknown = true) then is this project in danger of falling behind the Zendesk API without us knowing?
Also I'd suggest it should either be done for the whole system or not at all.

The balance is that with this annotation the code becomes more robust and users are not so dependent on making emergency fixes, especially when the release cadence is so slow.

Thoughts welcome

drd:
@gvh1234 it's a tricky balance to strike, and I'm not sure what is best for this library. However, FWIW:

```
$ git grep JsonIgnoreProperties
src/main/java/org/zendesk/client/v2/model/Collaborator.java:import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
src/main/java/org/zendesk/client/v2/model/Collaborator.java:@JsonIgnoreProperties(ignoreUnknown = true)
src/main/java/org/zendesk/client/v2/model/CustomFieldValue.java:import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
src/main/java/org/zendesk/client/v2/model/CustomFieldValue.java:@JsonIgnoreProperties(ignoreUnknown = true)
src/main/java/org/zendesk/client/v2/model/Macro.java:import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
src/main/java/org/zendesk/client/v2/model/Macro.java:@JsonIgnoreProperties(ignoreUnknown = true)
src/main/java/org/zendesk/client/v2/model/Photo.java:import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
src/main/java/org/zendesk/client/v2/model/Photo.java:@JsonIgnoreProperties(ignoreUnknown = true)
src/main/java/org/zendesk/client/v2/model/SatisfactionRating.java:import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
src/main/java/org/zendesk/client/v2/model/SatisfactionRating.java:@JsonIgnoreProperties(ignoreUnknown = true)
src/main/java/org/zendesk/client/v2/model/Ticket.java:import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
src/main/java/org/zendesk/client/v2/model/Ticket.java:@JsonIgnoreProperties(ignoreUnknown = true)
src/main/java/org/zendesk/client/v2/model/TicketForm.java:import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
src/main/java/org/zendesk/client/v2/model/TicketForm.java:@JsonIgnoreProperties(ignoreUnknown = true)
src/main/java/org/zendesk/client/v2/model/events/Event.java:import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
src/main/java/org/zendesk/client/v2/model/events/Event.java:@JsonIgnoreProperties(ignoreUnknown = true)
```

matthewtckr:
Personally, I'll take stability over being thorough with the Zendesk API, namely because a breaking change seems to always happen late on a Friday by Zendesk's developers, which means alerts all weekend as the ETL process fails unless I compile a new build.

I'm not opposed to having two different client libraries, if nothing else than to catch any new, "unknown" properties.  If there's a better way to do this, without causing an increased risk for production jobs, I'm all for it.

PR #187, opened 2017-08-08
Ticket Import

robert-fernandes:
Add integration with Ticket Import API -
 https://developer.zendesk.com/rest_api/docs/core/ticket_import


robert-fernandes:
This is not entirely correct.  The ticket object payload for import has a collection of tickets, not just a single ticket.  I will submit another pull request with that addition.

PR #188, opened 2017-08-09
Ticket Import

robert-fernandes:
Add support for Ticket Import API.

Extend Ticket to create a separate TicketImport object that holds a collection of Comments, not a single Comment.


robert-fernandes:
FYI, just completed using both importTicket and importTickets into a live instance, and things work as expected.  Hopefully this gets to make it into the official client someday, so I can stop managing my own fork.

Issue #189, opened 2017-08-24
use Boolean in models instead of boolean


abdelrahmanmohamed:
the default value of boolean is false, that means that the serializer will always serialize it to false in every request example:
```
Ticket x=new Ticket();
x.setId(ticket.getId());
```
json will be
```
{
  "tickets" : [ {
    "id" : 409671,
    "has_incidents" : false
  } ]
}
```
because of `private boolean hasIncidents;` need to be change to `private Boolean hasIncidents;` also the setter and the getters
i will create PR for it by the end of the day

abdelrahmanmohamed:
pull request created

PR #190, opened 2017-08-24
fix boolean to Boolean Issue #189

abdelrahmanmohamed:
the default value of boolean is false, that means that the serializer will always serialize it to false in every request example:
```
Ticket x=new Ticket();
x.setId(ticket.getId());
```
json will be
```
{
  "tickets" : [ {
    "id" : 409671,
    "has_incidents" : false
  } ]
}
```
because of `private boolean hasIncidents;` need to be change to `private Boolean hasIncidents;` also the setter and the getters

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,9 @@
     <developer>
       <name>Yoann Dubreuil</name>
     </developer>
+    <developer>
+      <name>Abdelrahman Sayed</name>
+    </developer>
   </developers>
 
   <mailingLists>
--- a/src/main/java/org/zendesk/client/v2/model/Brand.java
+++ b/src/main/java/org/zendesk/client/v2/model/Brand.java
@@ -17,13 +17,13 @@ public class Brand {
     @JsonProperty("brand_url")
     private String brandUrl;
     @JsonProperty("has_help_center")
-    private boolean hasHelpCenter;
+    private Boolean hasHelpCenter;
     @JsonProperty("help_center_state")
     private String helpCenterState;
     @JsonProperty("active")
-    private boolean isActive;
+    private Boolean isActive;
     @JsonProperty("default")
-    private boolean isDefault;
+    private Boolean isDefault;
     private Attachment logo;
     @JsonProperty("ticket_form_ids")
     private List<Long> ticketFormIds;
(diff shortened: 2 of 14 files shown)


aheritier:
@abdelrahmanmohamed you mean that by replacing boolean -> Boolean by default they are null thus not serialised any more ?
I imagine in that case using primitive types for long/int/... (if any) then we should have the same issue...
Our automated tests (which are a bit light) are ok with your change but this PR changes the API and it might break various consumers who weren't protected boolean against null checks.
wdyt @stephenc ?

abdelrahmanmohamed:
Yes, you will have the same issue if you are using primitive types int,long,double,byte ....etc and it will be serialised to zero, and that is the normal and expected behaviour, if you have any model that use primitive types it should be fixed by using wrapper classes unless the attribute have default business value

stephenc:
We can make breaking changes as this is a 0.x.y version... though it is better to stack them all up

pom.xml
```diff
@@ -52,6 +52,9 @@
     <developer>
       <name>Yoann Dubreuil</name>
     </developer>
+    <developer>
```
johnou: perhaps you mean contributor? :)

johnou:
@abdelrahmanmohamed can you give me some actual use cases of where these changes help?

johnou:
I'm not convinced, if you have a good few use cases for this pull request please reopen and comment.

abdelrahmanmohamed:
@johnou lets say you have this class
```
class Ticket{
     int TicketId;
     boolean isOpened;
     String email;
}
```
the value for isOpened is true in the server
and you want to update the email using this model so you want to only send the ticketId and the email fields but when the json serializer serialize  the model it will found a default value for the isOpened = false so it will serialize it and that will cause the isOpened to be updated  to false

johnou:
The isOpened field doesn't exist in ticket, I would rather do this on a case by case basis instead of converting all booleans to Booleans, probably even possible without a major release eg.

```java
public class Ticket extends Request implements SearchResultEntity {

    @JsonProperty("has_incidents")
    private Boolean hasIncidents;

    public boolean isHasIncidents() {
        return hasIncidents != null ? hasIncidents : false;
    }

    public void setHasIncidents(boolean hasIncidents) {
        this.hasIncidents = hasIncidents;
    }
}
```

PR #194, opened 2017-09-01
Overflow fix: Changed type of article attachment ID and article ID from int to Long.

ngyewch:
Another ID overflow fix.


johnou:
@ngyewch any particular reason why you used object long instead of primitive?

aheritier:
@johnou all classes under model are using Long for ids (probably to handle the JSON (de)serialization if the attribute doesn't exist) but Zendesk is using long ....

PR #195, opened 2017-09-01
Declare environment variables within the environment bloc (That's cleaner. Declarative pipeline rocks. Thanks @jglick)

aheritier:
Based on @jglick feedback to simplify the syntax
Note that this syntax doesn't support an environment with a dot (with is UGLY I admit), I'm not sure if this is something I should report cc @abayer 
FYI @stephenc

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,22 +1,22 @@
 pipeline {
+    environment {
+        ZENDESK_JAVA_CLIENT_TEST_URL             = credentials('ZENDESK_JAVA_CLIENT_TEST_URL')
+        ZENDESK_JAVA_CLIENT_TEST_USERNAME        = credentials('ZENDESK_JAVA_CLIENT_TEST_USERNAME')
+        ZENDESK_JAVA_CLIENT_TEST_PASSWORD        = credentials('ZENDESK_JAVA_CLIENT_TEST_PASSWORD')
+        ZENDESK_JAVA_CLIENT_TEST_TOKEN           = credentials('ZENDESK_JAVA_CLIENT_TEST_TOKEN')
+        ZENDESK_JAVA_CLIENT_TEST_REQUESTER_EMAIL = credentials('ZENDESK_JAVA_CLIENT_TEST_REQUESTER.EMAIL')
+        ZENDESK_JAVA_CLIENT_TEST_REQUESTER_NAME  = credentials('ZENDESK_JAVA_CLIENT_TEST_REQUESTER.NAME')
+    }
     agent {
         label "standard"
     }
     stages {
         stage("Build") {
             steps {
-                withCredentials(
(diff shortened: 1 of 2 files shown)


aheritier:
@stephenc on CI I prefer to use the env var instead of the property file which could be accessed from the workspace when the build is running

Jenkinsfile
```diff
@@ -1,22 +1,22 @@
 pipeline {
+    environment {
+        ZENDESK_JAVA_CLIENT_TEST_URL             = credentials('ZENDESK_JAVA_CLIENT_TEST_URL')
+        ZENDESK_JAVA_CLIENT_TEST_USERNAME        = credentials('ZENDESK_JAVA_CLIENT_TEST_USERNAME')
+        ZENDESK_JAVA_CLIENT_TEST_PASSWORD        = credentials('ZENDESK_JAVA_CLIENT_TEST_PASSWORD')
```
stephenc: probably not all of these should be credentials... and the username/password should be the same credentials not two separate SecretText credentials

stephenc:
@aheritier FYI have you tried creating an environment variable with a name including a `.` go on, give it a try, bash will not be your friend

screen shot 2017-09-01 at 16 10 47

aheritier:
good point @stephenc 
I just reused what was implemented by @matthewtckr in 4051855e56cb00932cb120fecdf27b437c4dec65

Jenkinsfile
aheritier: For now I created distinct entries to have the build working (feel free to edit
https://opensource.ci.cloudbees.com )

https://jenkins.io/doc/pipeline/tour/environment/#credentials-in-the-environment

I'm not sure we can control the naming of the generated environment variables if we use a Username/Password credential

stephenc: https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/e2f89933976a8a2885ca5b061ab3fb1d8a8fe7bb/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/model/CredentialsBindingHandler.java#L162-L165

A usernamepassword credential will be bound to three env vars:

`FOO` which gets username:password
`FOO_USR` which gets username
`FOO_PWD` which gets password

I would like some guidance from @abayer on how this code gets used though

aheritier: yep @stephenc this is what I understood. We just need to rename our properties to match this pattern usr/pwd instead of username/password
Not a problem from my POV

Jenkinsfile
```diff
@@ -1,22 +1,22 @@
 pipeline {
+    environment {
+        ZENDESK_JAVA_CLIENT_TEST_URL             = credentials('ZENDESK_JAVA_CLIENT_TEST_URL')
```
jglick: Need not be secret I guess?

aheritier: yes I prefer, it's our cloudbees sandbox env for now

PR #200, opened 2017-10-25
Added full test coverage for TemplateUri

kirksc1:


duemir:
@kirksc1 `testToString_whenDateValue_thenUriBuilt` test doesn't take into account TimeZones. Please review.

src/test/java/org/zendesk/client/v2/TemplateUriTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.text.ParseException;
+import java.util.*;
+
+import static org.junit.Assert.assertEquals;
+
+public class TemplateUriTest {
+
+    @Rule
+    public UTCRule utcRule = new UTCRule();
```
johnou: must be public.

johnou:
@aheritier ptal

PR #212, opened 2018-01-17
Add createTicketAsync method.

johnou:
Fixes https://github.com/cloudbees/zendesk-java-client/issues/211

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -273,10 +273,20 @@ public class Zendesk implements Closeable {
         complete(submit(req("DELETE", tmpl("/tickets/{id}.json").set("id", id)), handleStatus()));
     }
 
+    public ListenableFuture<JobStatus<Ticket>> createTicketDeferred(Ticket ticket) {
+        return submit(req("POST", cnst("/tickets.json?async=true"),
+                JSON, json(Collections.singletonMap("ticket", ticket))),
+                handleJobStatus(Ticket.class));
+    }
+
+    public ListenableFuture<Ticket> createTicketAsync(Ticket ticket) {
+        return submit(req("POST", cnst("/tickets.json"),
+                JSON, json(Collections.singletonMap("ticket", ticket))),
+                handle(Ticket.class, "ticket"));
+    }
+
     public Ticket createTicket(Ticket ticket) {
-        return complete(submit(req("POST", cnst("/tickets.json"),
-                        JSON, json(Collections.singletonMap("ticket", ticket))),
-                handle(Ticket.class, "ticket")));
+        return complete(createTicketAsync(ticket));
     }
 
     public JobStatus<Ticket> createTickets(Ticket... tickets) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -273,6 +273,12 @@ public void deleteTicket(long id) {
...
     }
 
+    public ListenableFuture<JobStatus<Ticket>> createTicketAsync(Ticket ticket) {
+        return submit(req("POST", cnst("/tickets.json?async=true"),
+                JSON, json(Collections.singletonMap("ticket", ticket))),
+                handleJobStatus(Ticket.class));
+    }
+
     public Ticket createTicket(Ticket ticket) {
         return complete(submit(req("POST", cnst("/tickets.json"),
```
bsideup: shouldn't it use `complete(createTicketAsync(ticket))` now? :)

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -273,6 +273,12 @@ public void deleteTicket(long id) {
         complete(submit(req("DELETE", tmpl("/tickets/{id}.json").set("id", id)), handleStatus()));
     }
 
+    public ListenableFuture<JobStatus<Ticket>> createTicketAsync(Ticket ticket) {
+        return submit(req("POST", cnst("/tickets.json?async=true"),
```
bsideup: Is it possible to create a ticket without `async=true` but with a `ListenableFuture` as a return type?

johnou: If you can come up with a good name for it so we can have both i'll put it right in ;)

bsideup: `ListenableFuture<JobStatus<Ticket>> createTicketAsync(Ticket ticket, boolean sendToBackground)` ? :)

johnou: one returns a job status, the other returns an actual ticket https://developer.zendesk.com/rest_api/docs/core/tickets#create-ticket

bsideup: `ListenableFuture<Ticket> createTicketAsync(Ticket ticket)`
`ListenableFuture<JobStatus<Ticket>> createTicketDeferred(Ticket ticket)`

? :)

johnou: I think that works I'll update the pr when I get back to the keyboard.

johnou:
@stephenc PTAL, I'm not completely happy with the method names as the deferred method uses a param called async..

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -273,10 +273,20 @@ public void deleteTicket(long id) {
         complete(submit(req("DELETE", tmpl("/tickets/{id}.json").set("id", id)), handleStatus()));
     }
 
+    public ListenableFuture<JobStatus<Ticket>> queueCreateTicketAsync(Ticket ticket) {
```
johnou: @bsideup ptal, using the same terminology Zendesk use in their documentation.

bsideup: 👍

johnou:
@aheritier PTAL

stephenc:
FTR I am fine with having `Async` on methods that return a future as long as we are consistent

PR #214, opened 2018-01-17
[Fixes #210] Because of the introduction of Multi-select custom field a value must be an array

aheritier:
https://support.zendesk.com/hc/en-us/articles/360000032087-New-ticket-field-type-Multi-select

PR #176, opened 2017-05-31
Update async http client to 2.0.38 version

mbyrdziak:

PR #215, opened 2018-01-18
Update async-http-client to 1.9.40

johnou:
(the transient Netty dep has some important fixes).


aheritier:
@johnou I think we should just merge #176 but for this we need to fix the master build with #214 WDYT ?

johnou:
@aheritier sure as long as we update it to latest 2.0.x before merging :+1:

PR #218, opened 2018-01-18
[WIP] Let's try to improve our tests and the coverage

aheritier:


johnou:
@aheritier what's the actual coverage? might be interesting to integrate coveralls or something.

aheritier:
I don't know what it is technically (not yet computed) but from my POV it is low. 
yes coveralls or maybe sonarsource, I think they allow OSS projects. I need to verify what we have to do

PR #238, opened 2018-02-07
async-http-client 2.3.0

johnou:
Important fixes in Netty, namely https://github.com/netty/netty/issues/7617


aheritier:
@duemir or @Dohbedoh noticed such kind of issue where the JVM didn't stop after downloading an attachment from ZD. Great work @johnou @slandelle @normanmaurer

Issue #204, opened 2017-11-27
bug: wrong URI creation with params


asafronienkov:
We have a query that looks like this, which fails:

zendeskClient.getSearchResults(Ticket.class, "requester:1111",                                                                 "sort_by=created_at&sort_order=desc&page=1&per_page=11");

This results in a query:
http://company.zendesk.com/api/v2/search.json?query=requester%3A1111%2Btype%3Aticket&params=sort_by%3Dcreated_at%2Bsort_order%3Ddesc%2Bpage%3D1%2Bper_page%3D11

which returns default results with page 100 instead of indicated 11.
so query should be smth like this  
query=type:ticket requester:1111&sort_by=created_at&sort_order=desc&page=2&per_page=11

Helmsdown:
So the implementation of ```org.zendesk.client.v2.Zendesk#getSearchResults(java.lang.Class<T>, java.lang.String, java.lang.String)``` is interesting. Was it intended that "params" be things like sort order? If so, should it have been a string? Or some Map<String, String>?

I need to perform a search and sort the results by updated_at. I'd like to submit a PR to support this and overcome the ```// TODO search with sort order``` comment in Zendesk.java 

Could a repo owner chime in here?

PR #240, opened 2018-02-07
Search Ordering

Helmsdown:
Related to #204 

I'd like some feedback on this PR. Specifically around testing. It seems like there is an opportunity to use wiremock to have unit tests which actually make http calls.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -40,6 +40,7 @@ import org.zendesk.client.v2.model.OrganizationField;
 import org.zendesk.client.v2.model.OrganizationMembership;
 import org.zendesk.client.v2.model.SatisfactionRating;
 import org.zendesk.client.v2.model.SearchResultEntity;
+import org.zendesk.client.v2.model.SortOrder;
 import org.zendesk.client.v2.model.Status;
 import org.zendesk.client.v2.model.SuspendedTicket;
 import org.zendesk.client.v2.model.Ticket;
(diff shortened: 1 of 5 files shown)


Helmsdown:
@aheritier would you approve of a new test class that uses wiremock to start testing new functionality and older functionality we as we get to it?

aheritier:
I think it should be better to have a new test class dedicated to these tests with Mocks
We keep the other with real integration tests

Helmsdown:
I agree. I did not mean to imply that the real integration tests should go away.

Helmsdown:
I am working on a sample test right now. Will push once its working.

Helmsdown:
@aheritier I ran into a problem. Wiremock delegates to ```java.net.URI#create``` which does not like certain characters. So a simple search url like the following blows up in Wiremock because of the space between search terms.

```
search.json?query=type:ticket status:closed&sort_by=status&sort_order=desc
```

https://github.com/tomakehurst/wiremock/issues/713 is open to fix this issue, but it's unclear when the author is going to get to it. 

What can we add to RealSmokeTest to give us some confidence about the PR?

Helmsdown:
@aheritier i figured out the problem with WIremock. Check out the test class I wrote. We could slowly duplicate tests for existing functionality and then each contributor can run them locally.

Helmsdown:
@aheritier can you circle back on this please.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1510,10 +1511,36 @@ public void deleteOrganizationMembership(long user_id, long organization_members
     }
 
     public <T extends SearchResultEntity> Iterable<T> getSearchResults(Class<T> type, String query) {
-        return getSearchResults(type, query, null);
+        return getSearchResults(type, query, Collections.emptyMap());
     }
 
+    /**
+     * Use {@link #getSearchResults(Class, String, Map)} instead.
```
alecharp: could be 

```
@deprecated Use {@link ...
```

Helmsdown:
Looks like this sat for about a year and a half. What do I need to do to get it over the finish line?

stephenc:
@aheritier

src/main/java/org/zendesk/client/v2/Zendesk.java
alecharp: Thinking about it we could keep both methods with deprecating one. This one, without a `SortOrder` could offer a default order.

aheritier: +1 I think it is what @PierreBtz did for Comment ordering

PR #248, opened 2018-04-04
Fixes #186 by updating the URI to the correct templated one.

### LINKED ISSUE #186: createSection() url is incorrect
createSection() method uses constant url /help_center/sections.json. But according to documentation it is required to specify category id help_center/categories/{id}/sections.json. Which leads to 404 error in response.

duemir:
The existing URI is replaced with `/help_center/categories/{id}/sections.json`
based on [Create Section][1] API documentation. It requires category,
hence the `categoryId` field of the `section` is made mandatory.

[1]: https://developer.zendesk.com/rest_api/docs/help_center/sections#create-section

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -1719,8 +1719,9 @@ public class Zendesk implements Closeable {
                 handleList(Translation.class, "translations"));
     }
     public Section createSection(Section section) {
-        return complete(submit(req("POST", cnst("/help_center/sections.json"), JSON,
-                json(Collections.singletonMap("section", section))), handle(Section.class, "section")));
+        checkHasCategoryId(section);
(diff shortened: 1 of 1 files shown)


duemir:
@aheritier we closed one of the tickets required for Integration Test, see
https://github.com/cloudbees/zendesk-java-client/commit/a00d2c9ca2042c3fc83b2ca4a8d3d95f17a281a7#diff-af7bc3f29194bb5b012e12a59fc5af96R214
https://cloudbees1433516384.zendesk.com/agent/tickets/22/events
It all looks fragile. Any ideas how we can improve it?

aheritier:
@duemir :( I will punish @MCardenasBlanco :-)
This is a real problem with integration tests.
Maybe in the title I should say DO NOT CLOSE
Otherwise the only solution is to create the required cases using the APIs before the test but if the creation fails ...

MCardenasBlanco:
@aheritier wut?

PR #255, opened 2018-06-04
int to long for Macro dto.

johnou:
org.zendesk.client.v2.ZendeskException: java.lang.IllegalArgumentException: Numeric value (360001871809) out of range of int (-2147483648 - 2147483647)
 at [Source: UNKNOWN; line: -1, column: -1]
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.zendesk.client.v2.model.Macro["id"])
	at org.zendesk.client.v2.Zendesk.complete(Zendesk.java:2199)
	at org.zendesk.client.v2.Zendesk.access$1400(Zendesk.java:93)

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Macro.java
+++ b/src/main/java/org/zendesk/client/v2/model/Macro.java
@@ -17,7 +17,7 @@ public class Macro implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
-    private int id;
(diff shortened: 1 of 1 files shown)


johnou:
@aheritier PTAL

johnou:
Been running this in production for a few hours now, unfortunately new macros caused bootstrap to fail..

PR #257, opened 2018-06-14
Adding of features to be GDPR compliant

danielzimmermann:
Added deletion of user, ticket and tickets:

- https://developer.zendesk.com/rest_api/docs/core/users#permanently-delete-user
- https://developer.zendesk.com/rest_api/docs/core/tickets#delete-tickets-permanently
- https://developer.zendesk.com/rest_api/docs/core/tickets#delete-multiple-tickets-permanently


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -267,15 +267,24 @@ public Ticket getTicket(long id) {
                 handleList(User.class, "users")));
     }
 
-    public void deleteTicket(Ticket ticket) {
```
johnou: might want to keep the old signature so it's backwards compatible?

danielzimmermann: thank you, i updated it, it was a copy/paste removal

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -267,6 +267,12 @@ public Ticket getTicket(long id) {
                 handleList(User.class, "users")));
     }
 
+    public JobStatus permanentlyDeleteTicket(Ticket ticket) {
+        checkHasId(ticket);
+        deleteTicket(ticket.getId());
```
johnou: deleteTicket is invoked twice, once here then again in permanentlyDeleteTicket.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -773,6 +797,17 @@ public void deleteUser(long id) {
         complete(submit(req("DELETE", tmpl("/users/{id}.json").set("id", id)), handleStatus()));
     }
 
+    public User permanentlyDeleteUser(User user) {
+        checkHasId(user);
+        deleteUser(user.getId());
```
johnou: Same issue here, double delete user.

src/test/java/org/zendesk/client/v2/RealSmokeTest.java
```diff
@@ -1,57 +1,24 @@
...
-import org.zendesk.client.v2.model.TicketForm;
-import org.zendesk.client.v2.model.User;
+import org.zendesk.client.v2.model.*;
 import org.zendesk.client.v2.model.events.Event;
-import org.zendesk.client.v2.model.hc.Article;
-import org.zendesk.client.v2.model.hc.Category;
-import org.zendesk.client.v2.model.hc.Section;
-import org.zendesk.client.v2.model.hc.Subscription;
-import org.zendesk.client.v2.model.hc.Translation;
+import org.zendesk.client.v2.model.hc.*;
```
johnou: Please revert and keep single imports.

johnou:
@danielzimmermann looks good to me, thanks!

johnou:
@danielzimmermann given this pull request is for GDPR, would you have time to also implement this endpoint? https://developer.zendesk.com/rest_api/docs/core/users#show-compliance-deletion-statuses

danielzimmermann:
@johnou i can do that: https://developer.zendesk.com/rest_api/docs/core/users#show-compliance-deletion-statuses but i need a couple of days for that, because i am a bit under stress right now :)

aheritier:
LGTM. Good work @danielzimmermann Thx a lot

Issue #262, opened 2018-07-30
Wrong typing for the 'value' field of Actions


Dohbedoh:
Retrieving triggers that have action to notify a target of type HTTP Target:

<img width="839" alt="http-target-with-parameters" src="https://user-images.githubusercontent.com/6858329/43412072-5e1f56a4-942c-11e8-978e-686feb7b446b.png">

fails with the following exception:

```
Exception in thread "main" org.zendesk.client.v2.ZendeskException: java.lang.IllegalArgumentException: Cannot deserialize instance of `java.lang.String` out of START_ARRAY token
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.zendesk.client.v2.model.Trigger["actions"]->java.util.ArrayList[0]->org.zendesk.client.v2.model.Action["value"]->java.lang.Object[][1])
        at org.zendesk.client.v2.Zendesk.complete(Zendesk.java:2239)
        at org.zendesk.client.v2.Zendesk.access$1400(Zendesk.java:94)
        at org.zendesk.client.v2.Zendesk$PagedIterable$PagedIterator.hasNext(Zendesk.java:2471)
        at com.cloudbees.zendesk.cli.admin.ListTriggersCmd.run(ListTriggersCmd.java:54)
        at com.cloudbees.zendesk.cli.CliApplication.run(CliApplication.java:416)
        at com.cloudbees.zendesk.cli.CliApplication.main(CliApplication.java:78)
Caused by: java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: Cannot deserialize instance of `java.lang.String` out of START_ARRAY token
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.zendesk.client.v2.model.Trigger["actions"]->java.util.ArrayList[0]->org.zendesk.client.v2.model.Action["value"]->java.lang.Object[][1])
        at java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:357)
        at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1895)
        at org.asynchttpclient.netty.NettyResponseFuture.get(NettyResponseFuture.java:202)
        at org.zendesk.client.v2.Zendesk.complete(Zendesk.java:2226)
        ... 5 more
Caused by: java.lang.IllegalArgumentException: Cannot deserialize instance of `java.lang.String` out of START_ARRAY token
[...]

PR #263, opened 2018-07-30
[#262] Fix Action model value type

Dohbedoh:
@reviewbybees


src/test/java/org/zendesk/client/v2/model/TriggerTest.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    private byte [] writeJson(Trigger trigger) throws JsonProcessingException {
+        return new ObjectMapper().writeValueAsBytes(trigger);
+    }
+
+    @Test
+    public void testTriggerWithComplexActionValues() throws Exception {
+        String json = "{\"id\":12345678,\"title\":\"Test Trigger\",\"active\":true,\"position\":45,\"conditions\":{\"all\":[],\"any\":[{\"field\":\"comment_includes_word\",\"operator\":\"includes\",\"value\":\"@triggerTest\"}]},\"actions\":[{\"field\":\"notification_target\",\"value\":[\"1234567890\",[[\"key1\",\"value1\"],[\"key2\",\"value2\"]]]}],\"createdAt\":null,\"updatedAt\":null}";
```
aheritier: @Dohbedoh Maybe it's like this in others tests but couldn't we move them to resources files and load them to have a better readability ?

src/main/java/org/zendesk/client/v2/model/Action.java
```diff
@@ -32,11 +32,11 @@ public void setField(String field) {
         this.field = field;
     }
 
-    public String[] getValue() {
+    public Object[] getValue() {
         return value;
     }
 
-    public void setValue(String[] value) {
+    public void setValue(Object[] value) {
```
duemir: The internal structure change is OK.
The problem, I believe, is changing the method signatures. The clients compiled against the older versions won't be able to use this new version because of class casting or method-binding issues if I understand correctly. 
This, therefore, might need a major version bump.

aheritier: In the past I used Clirr ( http://www.mojohaus.org/clirr-maven-plugin/ ) but I don't think it was updated to manage recent java versions. But for sure such tools are existing @duemir @Dohbedoh

Dohbedoh: Right I think this breaks compatibility. Although the signature change to a broader class... I am not sure at the moment how we could avoid this to fix that bug. Only solution I could think of is to use a custom Jackson serializer.

src/test/java/org/zendesk/client/v2/model/TriggerTest.java
Dohbedoh: Done. The compact `json` is used for the last assertion of each test.

src/main/java/org/zendesk/client/v2/model/Action.java
duemir: Arrays are covariant in Java so the setter is probably not an issue.
The getter is what worries me since in the client's code there'll be an assignment to a `String[]` but the new code returns `Object[]`

PierreBtz: I remember reading a solution for this kind of issue used by netbeans consisting of patching the bytecode

PR #270, opened 2018-09-25
Additional User Fields

matthewtckr:
Adding additional fields found in [Zendesk REST API Documentation](https://developer.zendesk.com/rest_api/docs/core/users)


matthewtckr:
I'm not a big fan of committing my own PRs, but it's been reviewed and approved for 30 days, so here it goes.

Thanks again @aheritier for the review!

PR #272, opened 2018-10-05
Finalize the migration to CodeShip by removing references to D@c

duemir:
* Links to D@c removed from Readme and pom.xml
* Temporary migration notes are removed
* Jenkinsfile is removed

@carlosrodlop are we ready to merge this? anything else I forgot to do?


duemir:
@carlosrodlop I have updated the PR as suggested so need a re-review

PR #276, opened 2018-10-17
adding more fields in Request

marydcouto:
I need access to set the request status to solved through api. It works only if solved field and its corresponding getters and setters are present in the Request. Create Request API was also failing as priority is mandatory field in the create request API. I also need access to couple of other fields, which I have added in the Request pojo.


marydcouto:
Please merge the changes, so I can integrate the zendesk request creation in my application. Thanks for approving the changes

marydcouto:
Could you please let me know how long will it take to get these changes merged? I am new to public github, so I am not familiar about how things work in public github.

aheritier:
Hi Mary,

  It will take some time. In open-source communities, maintainers aren't active every day and I would prefer to have others reviews because myself I'm not actively using it.
  Your changes are looking good when I compare with the documentation but without adding some new tests it is hard to guess if it will be completely fine. This part of the code has sadly a low test coverage : https://github.com/cloudbees/zendesk-java-client/blob/master/src/test/java/org/zendesk/client/v2/RealSmokeTest.java#L609
  After the validation of active others developers like @duemir or @johnou you'll have to also wait for a release to grab the new version from maven central.

HTH

src/main/java/org/zendesk/client/v2/model/Request.java
```diff
@@ -26,6 +29,10 @@
     protected Date createdAt;
     protected Date updatedAt;
     protected Comment comment;
+	protected boolean solved;
```
johnou: Please maintain project code formatting.

johnou: Better to use object boolean here or solved will always default to false and may inadvertently change the entity state.

marydcouto:
I worked on the review comments, but the PR is not getting updated. Can you tell me how to update the pull request?

marydcouto:
Changes are done. But the pull request is not getting updated. Could you please tell me what should I do?

marydcouto:
Pull request is now showing as updated. Sorry, this is the first time I am raising Pull request in public git hub, so I accidentally sent many comments regarding the pull request. Could you please review now?

johnou:
@aheritier looking better.

johnou:
@marydcouto can I assume you have tested these changes with your application?

marydcouto:
Yes I have tested this code. I made changes for Requests only. I can see 2 tests are failing for createOrganization and deleteOrganization. I am not sure why these tests are failing in my build. Do you know why these tests are failing in my build?

PR #284, opened 2019-04-17
Add search by labels.

dastultz:


aheritier:
@dastultz  Could you try to add some integration tests please ?
See RealSmokeTest.java
Just tell us what to configure in our sandbox to have them passing

dastultz:
Sorry for the delay, I plan to do this, please don't close the issue. Thanks.

dastultz:
Added unit tests. I did not run them! Documented article contents needed in sandbox.

src/test/java/org/zendesk/client/v2/RealSmokeTest.java
```diff
@@ -824,6 +827,47 @@ public void getArticles() throws Exception {
...
+            Article 1 with title "AllLabelOne" and label "AllLabelA"
+            Article 2 with title "AllLabelTwo" and labels "AllLabelA" and "AllLabelB"
+         When a search by labels "AllLabelA", "AllLabelB"
+         Then we get Article 2 but not Article 1
+            because Article 2 has both labels and Article 1 has only one
+         */
+        Iterable<Article> result = instance.getArticleFromAllLabels(Arrays.asList("AllLabelA", "AllLabelB"));
+        Set<String> actualTitles = extractTitles(result);
+        assertThat(actualTitles.size(), is(1));
+        assertThat(actualTitles, IsCollectionContaining.hasItems("AllLabelOne"));
```
aheritier: @dastultz it should be `AllLabelTwo` which has both labels (based on your scenario)?

```suggestion
        assertThat(actualTitles, IsCollectionContaining.hasItems("AllLabelTwo"));
```

dastultz: Dang it! I looked that over closely and still missed it. Fixed.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -403,6 +403,16 @@ public JobStatus permanentlyDeleteTickets(long id, long... ids) {
                 .set("query", searchTerm).set("section", sectionId), handleList(Article.class, "results"));
     }
 
+    public Iterable<Article> getArticleFromSomeLabels(List<String> labels) {
```
johnou: Naming seems a little off to me, especially with the fact singular is used for returning a collection. How about something like..

getArticlesWithAllLabels
searchArticlesWithLabels

dastultz: I don't understand what is meant by "get" vs "search" if anything. The part I was looking to make clear was that one method requires at least one label to be found, the other requires all labels to be found. IMO, "WithAll" is good, but the idea of "with at least one" is not captured in the name "searchArticlesWithLabels". Not sure if "search" implies such behavior. You objected to the use of singular for returning a collection, I agree with that, I followed the existing pattern established by the `getArticleFromSearch` method. I think `getArticlesFromSomeLabels` and `getArticlesFromAllLabels` best captures the concerns, though I don't like the apparent API inconsistency with the singular Article vs Articles.

johnou: How about getArticlesFromAllLabels and getArticlesFromAnyLabels

dastultz: Works for me, as long as everyone doesn't mind the inconsistency of Article plurality with `getArticleFromSearch`. Should I make the change, or will someone else do it?

johnou: Or introduce corrected method with deprecated annotation on the singular.

dastultz: Who is the "gatekeeper" of the public-facing API? I'm trying to add functionality but not dictate API design in general. I feel like someone more qualified should handle the naming issue. I'm happy with getArticlesFromAllLabels and getArticlesFromAnyLabels. I'm going to make that change and hope it goes through.

PR #287, opened 2019-06-10
[Fixes #286] Updates due to the rest of the SandBox

### LINKED ISSUE #286: Restore Integration Tests using the new Sandbox
We still don't have a specific zendesk instance for our test (no thanks @zendesk) and our CloudBees Zendesk sandbox was reset thus we have to reconfigure it
We have to repopulate some data and reconfigure our Codeship integration ( See #279 )

aheritier:


aheritier:
We cannot launch builds on PRs thus let's merge

PR #288, opened 2019-06-19
Update Zendesk.java

cavincla:
This is a fix because I have this problem creating the ticket:

org.zendesk.client.v2.ZendeskException: org.zendesk.client.v2.ZendeskException: java.lang.IllegalArgumentException: Can not deserialize instance of java.lang.String[] out of VALUE_STRING token
 at [Source: N/A; line: -1, column: -1] (through reference chain: org.zendesk.client.v2.model.Ticket["custom_fields"]->java.util.ArrayList[0]->org.zendesk.client.v2.model.CustomFieldValue["value"])

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -2450,6 +2450,8 @@ public class Zendesk implements Closeable {
         mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
         mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
         mapper.setDateFormat(new ISO8601DateFormat());
+	mapper.configure(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY, true);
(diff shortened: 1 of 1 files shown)


ClaudioBull:
I generate the issue simple creating a Ticket with a custom field compose
by only one value instead of String[] of values

final CustomFieldValue customField = new CustomFieldValue();
customField.setId(Long.valueOf(fieldName));
customField.setValue(new String[] {fieldValue});
customFields.add(customField);

Il giorno mer 19 giu 2019 alle ore 16:31 Arnaud Heritier <
notifications@github.com> ha scritto:

heidiwags:
I'm getting the same issue.  The problem though is - you can't/don't see this happen depending on how zendesk is configured.  For instance - we have a macro in our sandbox environment - and no issue, but in our production environment - same code results in same exception

aheritier:
@heidiwags yes or just to provide a method to configure it.

aheritier:
We could make the Zendesk class protected and extensible like asked in #306 with a method to override or we could perhaps have a method accepting a Consumer for the Mapping (I think we are in Java 8 min and it's possible)

PR #293, opened 2019-08-27
Support for Dynamic Content

hscabral:
Added support to Dynamic Content as per https://developer.zendesk.com/rest_api/docs/support/dynamic_content

- List and maintain Items
- List and Item Variants

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -53,8 +53,8 @@ import org.zendesk.client.v2.model.TwitterMonitor;
 import org.zendesk.client.v2.model.User;
 import org.zendesk.client.v2.model.UserField;
 import org.zendesk.client.v2.model.UserRelatedInfo;
-import org.zendesk.client.v2.model.dynamic.Item;
-import org.zendesk.client.v2.model.dynamic.Variant;
+import org.zendesk.client.v2.model.dynamic.DynamicContentItem;
+import org.zendesk.client.v2.model.dynamic.DynamicContentItemVariant;
 import org.zendesk.client.v2.model.hc.Article;
 import org.zendesk.client.v2.model.hc.ArticleAttachments;
 import org.zendesk.client.v2.model.hc.Category;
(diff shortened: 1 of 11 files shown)


aheritier:
@hscabral do you think it could be not too difficult to try to add some integration tests ?

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1609,6 +1611,68 @@ public SatisfactionRating createSatisfactionRating(Ticket ticket, SatisfactionRa
         return createSatisfactionRating(ticket.getId(), satisfactionRating);
     }
 
+    //////////////////////////////////////////////////////////////////////
+    // Action methods for Dynamic Content - Items and Variants
+    //////////////////////////////////////////////////////////////////////
+
+    public Iterable<Item> getItems() {
```
johnou: Little dangerous using getItemX if this is specific to dynamic content as it may clash with another feature.

hscabral:
> @hscabral do you think it could be not too difficult to try to add some integration tests ?

I will look into it.

src/main/java/org/zendesk/client/v2/Zendesk.java
hscabral: @johnou how do you mean clash? I don't remember seeing this as ambiguous, or anything using reflection that could affect using this.

johnou: In the sense Zendesk may introduce Article items.. and Item would be ambiguous, why not rename to DynamicContentItem?

johnou: I checked what Zendesk named it in the official SDK.. https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Resources/Core/DynamicContentItems.php

hscabral: @johnou got it, sounds good, I will make this change and push here.

hscabral:
I have performed the class name refactoring and added a simple iteration and fetch test.

src/main/java/org/zendesk/client/v2/model/dynamic/DynamicContentItem.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, url, name, placeholder, defaultLocaleId, outdated, createdAt, updatedAt, variants);
+    }
+
+    @Override
+    public String toString() {
+        return "{" +
```
johnou: class name?

src/main/java/org/zendesk/client/v2/model/dynamic/DynamicContentItemVariant.java
```diff
@@ -0,0 +1,220 @@
...
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, url, content, localeId, outdated, active, isDefault, createdAt, updatedAt);
+    }
+
+    @Override
+    public String toString() {
+        return "{" +
+            " id='" + getId() + "'" +
```
johnou: class name in tostring?

aheritier:
I will update the `toString()` @johnou

PR #296, opened 2019-11-13
[OPS-8322] Configure dependabot

aheritier:


aheritier:
I didn't remember that we had the PR builds available with Codeship.
Some tests are failing but it's not related to the PR

PR #314, opened 2019-12-18
Supplying a notify_subscribers property along POST:/articles.json

maxixcom:
https://developer.zendesk.com/rest_api/docs/help_center/articles#create-article

Supplying a notify_subscribers property with a value of false will prevent subscribers to the article from receiving an article creation email notification. This can be helpful when creating many articles at a time. Specify the property in the root of the JSON object, not in the "article" object.


pom.xml
```diff
@@ -25,7 +25,7 @@
 
   <groupId>com.cloudbees.thirdparty</groupId>
   <artifactId>zendesk-java-client</artifactId>
-  <version>0.11.1-SNAPSHOT</version>
+  <version>0.11.1-maxix-SNAPSHOT</version>
```
aheritier: I understand why you did it but it should be in the PR

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -55,12 +55,7 @@
 import org.zendesk.client.v2.model.UserRelatedInfo;
 import org.zendesk.client.v2.model.dynamic.DynamicContentItem;
 import org.zendesk.client.v2.model.dynamic.DynamicContentItemVariant;
-import org.zendesk.client.v2.model.hc.Article;
-import org.zendesk.client.v2.model.hc.ArticleAttachments;
-import org.zendesk.client.v2.model.hc.Category;
-import org.zendesk.client.v2.model.hc.Section;
-import org.zendesk.client.v2.model.hc.Subscription;
-import org.zendesk.client.v2.model.hc.Translation;
+import org.zendesk.client.v2.model.hc.*;
```
aheritier: I am not sure about the formatter configured in this project but you can probably discard such change and keep the full list of imports

maxixcom:
> Few comments but globally good.
> Take care I think #314 = #315 You are probably pushing to the same branch

Yes, I missed it unfortunately. Can we merge changes(#314,#315) from one branch. I can create new PR with both features. Will it work?

maxixcom:
New PR #316 was created with the feature

Issue #292, opened 2019-07-23
Add support for User Segments / Permissions Groups


Dohbedoh:
**Is your feature request related to a problem? Please describe.**

I cannot manage [User Segments](https://developer.zendesk.com/rest_api/docs/help_center/user_segments) / [Permissions](https://developer.zendesk.com/rest_api/docs/help_center/permission_groups) with the java client.

**Describe the solution you'd like**

An API to manage User Segments and Permissions Groups.

**Describe alternatives you've considered**

`curl` / `wget`

aheritier:
@Dohbedoh can you verify if #316 is properly fixing this issue please ?

PR #315, opened 2019-12-19
Help Center: Management Permission Groups

maxixcom:
https://developer.zendesk.com/rest_api/docs/help_center/permission_groups

A management permission group defines which agents can create, update, archive, and publish articles. It consists of a set of privileges, each of which is mapped to a user segment. Agents receive whichever privileges are associated with the user segments they belong to.

Should partly resolve: Add support for User Segments / Permissions Groups #292


maxixcom:
New PR #316 was created with the feature

PR #333, opened 2020-02-10
Create Locale.java

seraphgeorge:
Model for Locale


aheritier:
@seraphgeorge sorry for the delay, I missed this PR. The class looks good but should we add some methods to handle it ?

Issue #98, opened 2016-01-15
JobStatus related to createTickets doesn't seem to contain Ticket entities


jon-ruckwood:
In the case of `Zendesk#createTickets` it doesn't seem like `JobStatus#results` contains `Ticket` entities.

Given I want to create multiple tickets, and one or more of those tickets contains a validation error:

``` java
JobStatus<Ticket> jobStatus = zendesk.createTickets(tickets);
```

When I receive the JobStatus (or perform further `Zendesk#getJobStatus`), the JSON received is:

``` json
{
    "job_status": {
        "id": "1",
        "url": "https://subdomain.zendesk.com/api/v2/job_statuses/1.json",
        "total": 1,
        "progress": 0,
        "status": "failed",
        "message": "Failed at 2016-01-15 10:39:18 +0000",
        "results": [{
            "index": 0,
            "error": "TicketCreateFailed",
            "details": "Description: cannot be blank"
        }]
    }
}
```

But it's deserialized as (toString representation):

``` java
JobStatus{
  id='1',
  url='https://subdomain.zendesk.com/api/v2/job_statuses/1.json',
  total=1,
  progress=0,
  status=failed,
  message='Failed at 2016-01-15 10:39:18 +0000',
  results=[
    Ticket{
      assigneeId=null,
      id=null,
      url='null',
      externalId='null',
      type='null',
      subject='null',
      description='null',
      /* other properties omitted */
    }
  ]
}
```

Even when all the tickets are created successfully, deserialization of the `JobStatus` may fail as the JSON `$.job_status.results.status` value can be `Updated` which doesn't map to a `Ticket#status` enum value, e.g.:

``` json
"results":[
 {
    "id":4,
    "title":"Test 2 - ignore",
    "action":"update",
    "status":"Updated",
    "success":true,
    "errors":""
}
```

will cause:

```
java.lang.IllegalArgumentException: Can not construct instance of org.zendesk.client.v2.model.Status from String value 'Updated': value not one of declared Enum instance names: [new, open, pending, hold, solved, closed, deleted]
```

I'm not sure what the answer to this would be but I just wanted to make other people aware of this potential problem.

jon-ruckwood:
Also, I imagine this might be an issue for other API calls which return job statuses. (I noticed similar behaviour when attempting implementing the bulk update tickets endpoint).

wakingrufus:
I have run into this same issue when trying to implement this endpoint in the java client: 
[...]

Issue #96, opened 2016-01-14
Implementation of /api/v2/users/create_or_update_many.json


booshi:
This API does not have the implementation of the /api/v2/users/create_or_update_many.json method.

Accepts an array of up to 100 user objects. For each user, the user is created if it does not already exist, or the existing user is updated. It is painful to send the request to /create_many.json, parse thru the response and update each user in the list who already existed.

matthewtckr:
API Doc: https://developer.zendesk.com/rest_api/docs/core/users#create-or-update-many-users

matthewtckr:
Sample implementation available in matthewtckr/zendesk-java-client@a18bf05f063872e59646902351d33710620f051b commit (no smoke test)

ssteward54:
Are there any plans to support this functionality?

Issue #47, opened 2015-04-09
Implementation of /api/v2/users/update_many.json


unsync:
Hi,
Great job with this API, but there's no implementation of the /api/v2/users/update_many.json method.

This is greatly usefull to update a batch of users. Updating a batch of users one by one require a lot of thread to be opened.

Thanks in advance

wakingrufus:
I have opened a PR for this

wakingrufus:
My PR is here: https://github.com/cloudbees/zendesk-java-client/pull/260
my new test passed when I ran the tests locally. I am not sure why ci is failing so many tests....

any1one:
Hi !
My initial request was for the /users/update_many.json not /tickets/update_many.json, but thanks for your work ;)

wakingrufus:
oops! I will open a separate ticket for that. Sorry for the confusion!

wakingrufus:
@any1one Once I get update tickets merged, I will come back and see if I can implement this as well.

PR #345, opened 2020-05-29
Add various "Many" create/update/delete methods

aheritier:
Should solve: #47 #96  #98 #245

* New `deleteOrganizationMemberships` method
* New `createOrganizationMemberships` methods
* New `updateOrganizations` methods
* New `updateUsers` methods
* New `createOrUpdateUsers` methods
* [Breaking Change] `permanentlyDeleteTicket(s)` methods aren't calling anymore `deleteTicket(s)` because the deletion must be acked (and it's not necessarily immediate) before deleting them permanently
* [Breaking Change] New integration tests + **rework the model used to handle Async Jobs results** ( https://developer.zendesk.com/rest_api/docs/support/job_statuses#job-statuses )

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -1264,6 +1264,28 @@ public class Zendesk implements Closeable {
                                 organizationMembership))), handle(OrganizationMembership.class, "organization_membership")));
     }
 
+    /**
+     * https://developer.zendesk.com/rest_api/docs/support/organization_memberships#create-many-memberships
+     */
+    public JobStatus<OrganizationMembership> createOrganizationMemberships(OrganizationMembership... organizationMemberships) {
+        return createOrganizationMemberships(Arrays.asList(organizationMemberships));
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -801,14 +809,27 @@ public User createUser(User user) {
         return complete(createUsersAsync(users));
     }
 
+    public ListenableFuture<JobStatus<User>> createUsersAsync(List<User> users) {
```
aheritier, line 812: This one is moved here to match the organization of similar methods

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -335,6 +335,14 @@ public Ticket updateTicket(Ticket ticket) {
                 handle(Ticket.class, "ticket")));
     }
 
+    public JobStatus<Ticket> updateTickets(Ticket... tickets) {
```
aheritier, line 338: To follow the pattern used for similar methods

aheritier:
It's a draft, We have to add some tests

aheritier:
Replaced by #346 to have a build running on codeship

Issue #331, opened 2020-02-04
updatedAt() doesn't serialize properly in TicketImport


russellhoff:
**Describe the bug**
I've got a web service written in Spring Boot 2.2.4.RELEASE using Java 8. I can't create a ticket using Ticket Imports, throwing an exception `Invalid parameter: updated_at It must be an acceptable ISO 8601 date format`.

**To Reproduce**
```
Optional<Ticket> optTicket = zendesk.getTicket(pIdTicket);
if(optTicket.isPresent()){
  Ticket pTicket = optTicket.get();
  TicketImport ticketImport = new TicketImport();
  ticketImport.setCreatedAt(pTicket.getCreatedAt());
  // ...
  if(pTicket.getUpdatedAt()!=null)
    ticketImport.setUpdatedAt(pTicket.getUpdatedAt());
  // .. 
  Ticket duplicated =  zendesk.importTicket(ticketImport);
}
```

**Expected behavior**
The following exception is thrown:

```
2020-02-04 09:14:54.654 ERROR 17172 --- [nio-8080-exec-1] o.a.c.c.C.[.[.[.[dispatcherServlet]      : Servlet.service() for servlet [dispatcherServlet] in context with path [/zendesk_barik/api/prod/v2] threw exception [Request processing failed; nested exception is org.zendesk.client.v2.ZendeskResponseException: HTTP/400: Bad Request - {
  "error": {
    "title": "Atributo no válido",
    "message": "Pasó un valor no válido para el atributo updated_at. Invalid parameter: updated_at It must be an acceptable ISO 8601 date format"
  }
}
] with root cause

org.zendesk.client.v2.ZendeskResponseException: HTTP/400: Bad Request - {
  "error": {
    "title": "Atributo no válido",
    "message": "Pasó un valor no válido para el atributo updated_at. Invalid parameter: updated_at It must be an acceptable ISO 8601 date format"
  }
}

	at org.zendesk.client.v2.Zendesk$BasicAsyncCompletionHandler.onCompleted(Zendesk.java:2115) ~[zendesk-java-client-0.11.0.jar:?]
	at org.asynchttpclient.AsyncCompletionHandler.onCompleted(AsyncCompletionHandler.java:66) ~[async-http-client-2.10.4.jar:?]
	at org.asynchttpclient.netty.NettyResponseFuture.loadContent(NettyResponseFuture.java:222) ~[async-http-client-2.10.4.jar:?]
[...]

Issue #94, opened 2016-01-05
Ticket#dueAt is serialised incorrectly


jon-ruckwood:
When a Ticket is serialised it seems that `dueAt` is represented by a `long` instead of an ISO 8601 formatted String. This causes the API to reject the request.

Actual:

``` json
{"ticket":{"type":"task","has_incidents":false,"due_at":1420070462000}}
```

Expected:

``` json
{"ticket":{"type":"task","has_incidents":false,"due_at":"2015-01-01T00:01:02.000Z"}}
```

I fixed this in a fork (see: https://github.com/jon-ruckwood/zendesk-java-client/commit/3f112393f53a6e5a5f67d9a1b9b5d1b6b855623f), I'm not sure if that's useful at all.

mttdbrd:
I just came across this too. There are several ways to fix it. I'm not sure that @jon-ruckwood's fix is the cleanest.  There are some easier ways documented here (including using annotation directives to Jackson): http://stackoverflow.com/a/31613581/3022173

jon-ruckwood:
@mttdbrd +1 Yep it's cleaner fix. 

I was thinking that other Date fields across the model would need this serialisation applying also.

aheritier:
This is the problem to use old Date Objects. they are serialized by default to the timestamp
Replacing all Dates by OffsetDateTime should be better but it's a big API change :(
Changing globally the option like  @jon-ruckwood did is a good option if all date fields are managed like this (it seems to be the same than #331) but we can locally do it on each date field with JsonFormat like proposed by @mttdbrd

aheritier:
For now we have 

```
        mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
        mapper.setDateFormat(new StdDateFormat());
```

aheritier:
With #348 I cannot reproduce the issue in a test

PR #348, opened 2020-05-31
Validate dates management

aheritier:
* New test to validate #94 issue
* New tests to validate #331 issue

### DIFF
--- a/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
+++ b/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
@@ -1,5 +1,6 @@
 package org.zendesk.client.v2;
 
+import org.apache.commons.lang3.time.DateUtils;
 import org.hamcrest.Matchers;
 import org.hamcrest.core.IsCollectionContaining;
 import org.junit.After;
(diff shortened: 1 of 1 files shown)


src/test/java/org/zendesk/client/v2/RealSmokeTest.java
```diff
@@ -1669,6 +1786,38 @@ private Ticket newTestTicket() {
...
+        return ticketImports.toArray(new TicketImport[0]);
+    }
+
+    /**
+     * Creates a new ticketImport
+     */
+    private TicketImport newTestTicketImport() {
+        assumeThat("Must have a requester email", config.getProperty("requester.email"), notNullValue());
+        assumeThat("Must have a requester name", config.getProperty("requester.name"), notNullValue());
+        Date now = Calendar.getInstance().getTime();
```
PierreBtz, line 1806: Is it ok to use now?

aheritier: What do you mean ? The name of the variable ?

aheritier: yes it's because it's an integration test

PR #369, opened 2020-11-04
Fixes #368

### LINKED ISSUE #368: Wrong JsonProperty for Macro updatedAt field
**Describe the bug**
The JsonProperty on the method getUpdatedAt() for the class org.zendesk.client.v2.model.Macro is "deleted_at". This is not a valid json value [Zendesk spec](https://developer.zendesk.com/rest_api/docs/support/macros), which causes the field to be null.

Extract from Macro class (line 70-73):
`
    @JsonProperty("deleted_at")
`
`
    public Date getUpdatedAt() {
`
`
        return updatedAt;
 `
`
   }
`

**Expected behavior**
The JsonProperty should be "updated_at"

jellevangompel:


aheritier (approved):
LGTM +1 (probably a bad copy/paste)

aheritier:
(The build is failing, it's a known issue with external PRs) 
I verified locally the build is passing - which is a problem and is showing the origin of the issue: we don't test this code

jellevangompel:
@aheritier Thanks for the info. Do I need to do anything to get the pull request merged?

aheritier:
Let's wait a review by @duemir and we can merge / release it

duemir (approved):

PR #384, opened 2021-04-22
Add support for recovering suspended tickets

yusuke:
https://developer.zendesk.com/rest_api/docs/support/suspended_tickets#recover-multiple-suspended-tickets

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -93,6 +93,8 @@ import java.util.Objects;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 /**
  * @author stephenc
@@ -611,6 +613,15 @@ public class Zendesk implements Closeable {
     public Iterable<SuspendedTicket> getSuspendedTickets() {
         return new PagedIterable<>(cnst("/suspended_tickets.json"),
                 handleList(SuspendedTicket.class, "suspended_tickets"));
+    }
+
+    public Iterable<Ticket> recoverSuspendedTickets(List<SuspendedTicket> tickets) {
+        List<Long> ids = new ArrayList<>();
+        for (SuspendedTicket ticket : tickets) {
+            ids.add(ticket.getId());
+        }
+        return complete(submit(req("PUT", tmpl("/suspended_tickets/recover_many.json{?ids}").set("ids", ids)),
+                handleList(Ticket.class, "tickets")));
     }
 
     public void deleteSuspendedTicket(SuspendedTicket ticket) {


duemir (changes requested):
According to the docs

>Accepts up to 100 ticket ids. Note that suspended tickets that fail to be recovered are still included in the response.

It is not documented how API is going to fail, though. We should probably enforce this here somehow. WDYT @aheritier ?

aheritier:
@duemir logically the API will return a 400 error (but with Zendesk :-) )
I don't think we do such control in others APIs  (but not sure if we have some with such constraints)
Ideally writing some integration tests could help to validate the behavior, especially on limits but they are so painful to write (only us at cloudbees can use our test instance to design them)

aheritier:
(I have also to understand why the ITs are failing. I thought we skipped them in the past for PRs not coming from the main repo

yusuke:
Confirmed that the endpoint returns 422 Unprocessable Entity for such requests.
How about avoiding actual endpoint call, and throwing IllegalArgumentException instead when the number of ids exceeds 100?

```
2021-05-04 14:56:52,907 [main] DEBUG org.asynchttpclient.netty.request.NettyRequestSender - Using open Channel [id: 0x8f17a1fa, L:/192.168.1.115:50096 - R:samuraism.zendesk.com/104.16.51.111:443] for PUT '/api/v2/suspended_tickets/recover_many.json?ids=[snip]'
2021-05-04 14:56:54,143 [AsyncHttpClient-3-1] DEBUG org.asynchttpclient.netty.handler.HttpHandler - 

Request DefaultFullHttpRequest(decodeResult: success, version: HTTP/1.1, content: EmptyByteBufBE)
PUT /api/v2/suspended_tickets/recover_many.json?ids=[snip] HTTP/1.1
cookie: __cfduid=d320d1e9dd4bef27de7d6002f8d004bcf1620107769; __cfruid=35a4134579789c40f14b84ba64e60bd9946a02e3-1620107769
content-length: 0
host: samuraism.zendesk.com
authorization: Basic *****==
accept: */*
user-agent: AHC/2.1

Response DefaultHttpResponse(decodeResult: success, version: HTTP/1.1)
HTTP/1.1 422 Unprocessable Entity
Date: Tue, 04 May 2021 05:56:54 GMT
Content-Type: application/json; charset=UTF-8
Content-Length: 407449
Connection: keep-alive
X-Zendesk-API-Version: v2
X-Zendesk-Application-Version: v8356.114
X-Frame-Options: SAMEORIGIN
X-Rate-Limit: 200
X-Rate-Limit-Remaining: 188
Strict-Transport-Security: max-age=31536000;
Cache-Control: no-cache
X-Zendesk-Origin-Server: classic-app-server-599c955c74-kwchr
X-Request-Id: 649f7d06de31efaa-NRT
X-Runtime: 1.057104
X-Zendesk-Zorg: yes
X-Request-ID: 649f7d06de31efaa-NRT
CF-Cache-Status: DYNAMIC
cf-request-id: 09d78c784b0000efaadc326000000001
Expect-CT: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
Report-To: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report?s=Ov4p400zWENVQE9Qg03CsnJilFCx%2BT1xi9nQrUZnhXS15h7bll3LakYFWuzqY7FBhDVpavsOpboHmpbA%2FuEtu7X0xfgpkhs63SWe43Iw%2FBwUConQ7dU%3D"}],"max_age":604800,"group":"cf-nel"}
NEL: {"max_age":604800,"report_to":"cf-nel"}
Server: cloudflare
CF-RAY: 649f7d06de31efaa-NRT
```

yusuke:
Added a commit to throw IllegalArgumentException. I also added a JavaDoc comment to clarify the behavior.

PR #392, opened 2021-07-12
Add end_user_conditions to ticket form

internetstaff:
We needed end_user_conditions from the Ticket Forms as part of an internal project. 

I tried to mimic code style in this file, but i seemed a little weird. lmk.

Thanks!


duemir:
Documented in [Ticket Forms | Zendesk Developer Docs](https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_forms/).

duemir:
The structure of objects in the array is specified only by example but it appears to match what was defined here.
The new classes need getters and setters the same as enclosing class, otherwise these fields won't be accessible since they are all `private`

internetstaff:
Updated with getters/setters. 

My excuse is my use case is strange - I serialize these straight back to JSON. :)

PR #394, opened 2021-09-20
feat(searchPagination): add pagination search api for tickets

naveenreddyalka:


aheritier:
Thanks a lot for your contribution @naveenreddyalka 
I quickly scanned your PR and would like to give you some high level feedbacks:
* There is a lot of noise because of the introduction of `final` keywords almost everywhere. It makes the review hard to do and it's out of the scope of your change
* Don't change the Maven pom version in your PR (you can do it locally but not in the PR)
* PR tests are failing but it's maybe not your fault. It's because the external contributors cannot use our test environment

naveenreddyalka:
> Thanks a lot for your contribution @naveenreddyalka
> I quickly scanned your PR and would like to give you some high level feedbacks:
> 
> * There is a lot of noise because of the introduction of `final` keywords almost everywhere. It makes the review hard to do and it's out of the scope of your change
> * Don't change the Maven pom version in your PR (you can do it locally but not in the PR)
> * PR tests are failing but it's maybe not your fault. It's because the external contributors cannot use our test environment

Some of my local settings implied those changes but now I cleaned them up

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1749,7 +1753,82 @@ public void deleteOrganizationMembership(long user_id, long organization_members
...
+      final StringBuilder uriTemplate = new StringBuilder("/search.json{?query");
+
+      for (final String paramName : paramsMap.keySet()) {
+          uriTemplate.append(",")
+                  .append(paramName);
+      }
+
+      uriTemplate.append("}");
+      
+      String typeName = null;
```
romenrg, line 1797: Actually, this loop could also be extracted and shared with `getSearchResults`, since the only difference is `searchType.equals(entry.getValue())` vs `type.equals(entry.getValue())` and that var could be a parameter of the extracted private method

romenrg (commented):
Finally had a chance to review this. Thanks @naveenreddyalka for your contribution and @aheritier for your replies to our questions that led to this contribution

A few comments:

@naveenreddyalka, if I understood correctly, what you are doing here is to allow pagination by keeping the pagination logic out of the library, delegating its control to the app that passes the query to the library (which should keep track of the "pointers" from the results and make as many call as necessary to `getSearchTicketPageResults `). Did I understand that right? Seems like a valid enhancement to me, to provide an alternative to the existing implementation which seemed to always retrieve all the results (not letting the user handle the pagination).

As my two cents, I noticed most of the logic in `getSearchPageResults` is shared with the pre-existing `getSearchResults` (the `uriTemplate` entirely and the `typeName` loop if parametrized as explained in the comment I made in the corresponding line)

romenrg (approved):

aheritier:
The PR looks very clean now. thanks a lot @naveenreddyalka and @romenrg 
I will try to review it after this week and especially to validate our tests (they aren't working for external committers - it's painful)

aheritier:
@naveenreddyalka could you rebase your PR please ?
I fixed the build on the master branch (your PR should be validated after that)

naveenreddyalka:
> @naveenreddyalka could you rebase your PR please ?
> I fixed the build on the master branch (your PR should be validated after that)

Thanks, rebase done.

PR #415, opened 2021-11-08
Add type attribute to Comment model

besbes:
This adds the `type` attribute to the `Comment` model.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Comment.java
+++ b/src/main/java/org/zendesk/client/v2/model/Comment.java
@@ -23,6 +23,8 @@ public class Comment implements Serializable {
     private List<Attachment> attachments;
     private Date createdAt;
     private Boolean publicComment;
+    private CommentType type;
+
 
     public Comment() {
     }
@@ -102,6 +104,15 @@ public class Comment implements Serializable {
 
     public void setPublic(Boolean isPublic) {
         this.publicComment = isPublic;
+    }
+
+    @JsonProperty("type")
+    public CommentType getType() {
+        return type;
+    }
+
+    public void setType(CommentType type) {
+        this.type = type;
     }
 
     @Override
--- a/src/main/java/org/zendesk/client/v2/model/CommentType.java
+++ b/src/main/java/org/zendesk/client/v2/model/CommentType.java
@@ -1,0 +1,13 @@
+package org.zendesk.client.v2.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_comments/
+ */
+public enum CommentType {
+
+    @JsonProperty("Comment") COMMENT,
+    @JsonProperty("VoiceComment") VOICE_COMMENT;
+
+}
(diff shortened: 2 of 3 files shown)


aheritier (approved):
LGTM
I have to check if we can add an integration test

besbes:
@aheritier Thanks for the quick feedback. Let me know if I can contribute to the testing part — I didn't find an existing match so I wasn't sure how much coverage you want to have here.

aheritier:
when possible @besbes I try to verify that we can serialize/deserialize the objects but yes I am not sure we have some tests in ITs for Comments
Our ITs are failing, I also have to fix them. For some reason we now get a too many requests error

besbes:
@aheritier Let me know how I can support here. I couldn't find instructions how to run integration tests, what's the best way to check that?

aheritier:
Hi @besbes 

Did you test the code with a real zendesk instance?

I get an error

```
Caused by: java.lang.IllegalArgumentException: Cannot deserialize value of type `org.zendesk.client.v2.model.CommentType` from String "Comment": not one of the values accepted for Enum class: [VOICE_COMMENT, COMMENT]
 at [Source: UNKNOWN; byte offset: #UNKNOWN] (through reference chain: org.zendesk.client.v2.model.Comment["type"])
```

I updated your code and modified some tests to validate it.

could you have a look at it please ?

besbes:
@aheritier Yes I did (although not within this project because I didn't run integration tests). According to the [JsonProperty javadoc](https://javadoc.io/doc/com.fasterxml.jackson.core/jackson-annotations/latest/com/fasterxml/jackson/annotation/JsonProperty.html) this should work, so I'm not sure why it didn't for you? Happy to look into it — please just let me know how you run integration tests so I can try to reproduce.

aheritier:
@besbes I just added more details in #419 
Does it help ?

besbes:
@aheritier Yes, thank you. I found the issue — it looks like `@JsonProperty` on `enum` instances is not compatible with `DeserializationFeature.READ_ENUMS_USING_TO_STRING` (which we are using here). Therefore your change (using `toString()` to return the value for serialization makes sense. Thanks for implementing the integration test! Nothing to add from my side.

aheritier:
Hi @besbes

Ok it makes sense. Not sure why it was done like this (this project is old).

I'll wait for a review from @PierreBtz , @Dohbedoh or @duemir before merging it and publishing a release

aheritier (approved):

PierreBtz (changes requested):
Something looks odd, the Comment object is only describing the `COMMENT` type, the `VOICE_COMMENT` type does not use the same model as per https://developer.zendesk.com/documentation/ticketing/managing-tickets/adding-voice-comments-to-tickets/.

As far as I understand the documentation, a `VOICE_COMMENT` would not properly deserialize with this code.

aheritier:
@PierreBtz The specific JSON is only to create a comment of type Voice AFAIU. To read it it is the same, even if I am not sure how you get the URL and specific data for a voice comment

PierreBtz (approved):
Ok, if we agree that this code can only work in "read only" mode then lgtm.

Dohbedoh (approved):
Right. Though there is a value for `VOICE_COMMENT`, the current model does not allow to use it yet. But seems fine by me for now.

PR #416, opened 2021-11-12
Configure the polling interval/delay to be less aggressive in ITs

aheritier:
It avoids APIRateLimitExceeded errors

### DIFF
--- a/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
+++ b/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
@@ -1877,7 +1877,7 @@ public class RealSmokeTest {
         assertNotNull(result.getId());
         assertNotNull(result.getStatus());
 
-        // Let's wait for its completion (30 seconds max)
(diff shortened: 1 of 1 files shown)


src/test/java/org/zendesk/client/v2/RealSmokeTest.java
```diff
@@ -1874,8 +1877,8 @@ private JobStatus waitJobCompletion(final JobStatus result) {
         assertNotNull(result.getId());
         assertNotNull(result.getStatus());
 
-        // Let's wait for its completion (5 seconds max)
-        await().atMost(10, TimeUnit.SECONDS).until(() ->
+        // Let's wait for its completion (30 seconds max)
```
PierreBtz, line 1880:
```suggestion
        // Let's wait for its completion (2 minutes max)
```

PierreBtz (dismissed):
lgtm, except for a duration in a comment I'm not sure I get

Dohbedoh (approved):

PR #417, opened 2021-11-12
Allow to delete TicketForm

aheritier:
And use it in ITs to delete the created ones (we reached the limit of 300 forms)


PierreBtz (approved):
lgtm 🐝

Dohbedoh (approved):

PR #426, opened 2022-01-02
Add update Ticket Form endpoint

nkharabaruk:
- Added missing **PUT** ticket form endpoint
- changed `isEmptyString` -> `isEmptyOrNullString` in test config check because it was possible to have `null` when the property is missing


PierreBtz (approved):
lgtm, thanks for the contribution!

aheritier (approved):
Thanks a lot @nkharabaruk for this contribution
It's clean, the code is ok, new integration tests are relevant and passing (I verified locally)
All good to me (we should schedule to do a release this sprint @PierreBtz )

There are some tests failure but not related to the PR (job completions are taking more than 30sec - I will do a PR to increase the timeout to 1min which is allowing them to pass)

PR #427, opened 2022-01-03
Increate ITs timeout to 60s

aheritier:
They are failing on CI and locally with 30s

It should fix our ITs which are executed and failing on master


PierreBtz (approved):
lgtm 🐝

PR #429, opened 2022-01-06
[bug] fix an issue preventing searching tickets with email aliases (+)

PierreBtz:


aheritier (approved):
It looks good to me after the discussion we had together
You have a code formatting issue in RealSmokeTest otherwise it's good

aheritier (approved):
The change is ok for me, but the mass reformat in the test class is painful.
About this change when using "+" in searches we'll have to take care to highlight it in the release note (not sure how it will impact the users doing some searches - maybe they did some workarounds we didn't think about)

PR #458, opened 2022-03-06
Make Metric/ZendeskComboMinutes serializable (useful for caching)

### LINKED ISSUE #457: Add getUsers(long ids..) method
Should be able to get more than 1 user in a single request

codeconsole:
fixes #457

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Metric.java
+++ b/src/main/java/org/zendesk/client/v2/model/Metric.java
@@ -10,7 +10,7 @@ import java.util.Date;
  */
 public class Metric implements Serializable {
 
-    private static final long serialVersionUID = 1L;
+    private static final long serialVersionUID = -847290591071406141L;
 
     @JsonProperty("id")
     protected Long id;
@@ -247,7 +247,7 @@ public class Metric implements Serializable {
                 '}';
     }
 
-    public class ZendeskComboMinutes {
+    public class ZendeskComboMinutes implements Serializable {
 
         @JsonProperty("calendar")
         protected Long calendarMinutes;
--- a/src/main/java/org/zendesk/client/v2/model/Metric.java
+++ b/src/main/java/org/zendesk/client/v2/model/Metric.java
@@ -249,6 +249,8 @@ public class Metric implements Serializable {
 
     public class ZendeskComboMinutes implements Serializable {
 
+        private static final long serialVersionUID = 295719113530420878L;
+
         @JsonProperty("calendar")
         protected Long calendarMinutes;
         @JsonProperty("business")


aheritier (dismissed):
Thanks for your contribution @codeconsole 

LGTM, to be perfect I would I loved to:

* have 2 distinct commits (1 for the new method, 1 for the Serialization)
* have an integration test to validate the new method

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -854,6 +854,11 @@ public User getUser(long id) {
         return complete(submit(req("GET", tmpl("/users/{id}.json").set("id", id)), handle(User.class, "user")));
     }
 
+    public List<User> getUsers(long id, long... ids) {
+        return complete(submit(req("GET", tmpl("/users/show_many.json{?ids}").set("ids", idArray(id, ids))),
+                handleList(User.class, "users")));
+    }
```
PierreBtz, line 860:
```suggestion
    }
```

PierreBtz (changes requested):
Thanks for the contribution!

I think we need to make the `serialVersionUID` explicit instead of relying on the runtime to compute it.

If I was to nitpick, as Arnaud, I would prefer seeing two different commits since the two changes are not linked.

gareth-hughes-cloudpay:
Can I make a suggestion please?
This is exactly the query I was about to add myself but I need the variation
`/users/show_many.json?external_ids`
Could you please add this version of the method as well?

codeconsole:
> Thanks for your contribution @codeconsole
> 
> LGTM, to be perfect I would I loved to:
> 
> * have 2 distinct commits (1 for the new method, 1 for the Serialization)
> * have an integration test to validate the new method

All I did was copy the this method `List<Topic> getTopics(long id, long... ids)`
Where is the Integration test for that method?

I added the Serializable because basically I couldn't use this API at all due to the results were not cacheable. 
If I don't cache results, I get a `HTTP/429: Too Many Requests - Number of allowed API requests per minute exceeded` error.

Unfortunately via the API, I don't see a way of retrieving the requester's email with the Ticket, so I have to do a subseuquent request for User information.

It makes me wonder why Ticket.requester exists because I don't see it being set anywhere.

PierreBtz:
@codeconsole 

> Where is the Integration test for that method?

It does not exist, but you can take example on this one:

https://github.com/cloudbees-oss/zendesk-java-client/blob/8a045fdc5bda665ef2cfb33b445027348c7ccbcc/src/main/java/org/zendesk/client/v2/Zendesk.java

Tested here: https://github.com/cloudbees-oss/zendesk-java-client/blob/d9d7e01244850419412115df60e273d73ed4a1c8/src/test/java/org/zendesk/client/v2/RealSmokeTest.java#L1007-L1016

> I added the Serializable because basically I couldn't use this API at all due to the results were not cacheable.

It's ok to add it, just not in the same commit. Thanks for defining it explicitely btw :)

> It makes me wonder why Ticket.requester exists because I don't see it being set anywhere.

From what I understand reading the documentation, this field is useful if you are to create anonymous ticket requests (see https://developer.zendesk.com/api-reference/ticketing/tickets/ticket-requests/#additional-properties and https://developer.zendesk.com/documentation/ticketing/managing-tickets/creating-and-managing-requests/#creating-anonymous-requests).

The doc is not 100% for non anonymous requests but I think in this case you would use the requester id only and not the requester object.

PierreBtz:
@gareth-hughes-cloudpay I created a new task out of this request.

aheritier (approved):

PR #510, opened 2022-09-26
Add Automatic-Module-Name to manifest

oyarzun:
Use maven-jar-plugin to add `Automatic-Module-Name: org.zendesk.client.v2` to the manifest. This makes it easier to use with Java 9+ in a modular projects


PierreBtz (approved):
lgtm 🐝 thanks!

PR #521, opened 2022-10-25
add Via to Comment class according to Zendesk API spec

venediktov:
The Comment class is missing Via member , which allows detecting via what channel a ticket's comment was created.
Besides filtering comments based on "public": false or "public":true attribute our organization and perhaps others may need to utilize other attributes of Comment class which is available in json payload in Zendesk API but was omitted in this java library. 

Because the unit tests are more like integration tests in this project I wasn't able to run those tests yet, however if you require me to verify I can.

Let me know what you think.


PierreBtz (approved):
lgtm, thanks for the contribution!

I took the liberty of adding the via comment in the `toString` method, hope that's ok with you.

venediktov:
Thank you @PierreBtz for PR acceptance!

PR #545, opened 2023-03-02
implemented getCategoryTranslation, getSectionTranslation and getArticleTranslation.

Walti91:
added methods to fetch a single translation for a category, section or an article. as seen here https://developer.zendesk.com/api-reference/help_center/help-center-api/translations/#show-translation

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -2242,7 +2242,7 @@ public class Zendesk implements Closeable {
                 handleList(Translation.class, "translations"));
     }
 
-    public Translation getArticleTranslation(long articleId, String locale) {
+    public Translation showArticleTranslation(long articleId, String locale) {
         return complete(submit(req("GET",
                 tmpl("/help_center/articles/{articleId}/translations/{locale}.json")
                     .set("articleId", articleId).set("locale", locale)),
@@ -2341,7 +2341,7 @@ public class Zendesk implements Closeable {
                 handleList(Translation.class, "translations"));
     }
 
-    public Translation getCategoryTranslation(long categoryId, String locale) {
+    public Translation showCategoryTranslation(long categoryId, String locale) {
         return complete(submit(req("GET",
                 tmpl("/help_center/categories/{categoryId}/translations/{locale}.json")
                     .set("categoryId", categoryId).set("locale", locale)),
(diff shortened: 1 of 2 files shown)


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -2333,6 +2340,14 @@ public Iterable<Translation> getCategoryTranslations(Long categoryId) {
                 tmpl("/help_center/categories/{categoryId}/translations.json").set("categoryId", categoryId),
                 handleList(Translation.class, "translations"));
     }
+
+    public Translation getCategoryTranslation(long categoryId, String locale) {
```
PierreBtz, line 2344: I don't see any endpoint to show translations for Category, am I missing something?

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -2242,6 +2242,13 @@ public Iterable<Translation> getArticleTranslations(Long articleId) {
                 handleList(Translation.class, "translations"));
     }
 
+    public Translation getArticleTranslation(long articleId, String locale) {
```
PierreBtz, line 2245: Shouldn't we use the same verb as the Zendesk API: https://developer.zendesk.com/api-reference/help_center/help-center-api/translations/#show-translation:

```suggestion
    public Translation showArticleTranslation(long articleId, String locale) {
```

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -2384,6 +2399,14 @@ public Iterable<Translation> getSectionTranslations(Long sectionId) {
                 tmpl("/help_center/sections/{sectionId}/translations.json").set("sectionId", sectionId),
                 handleList(Translation.class, "translations"));
     }
+
+    public Translation getSectionTranslation(long sectionId, String locale) {
```
PierreBtz, line 2403: I don't see any endpoint to show translations for Section either.

PierreBtz (changes requested):
Thanks for the contribution!

We need to sort out if the endpoint you are using for category and section exist.

We will also need to add some tests to cover the new method(s). I know that it's a bit hard to write those tests since external contributors don't have access to our Zendesk sandbox. In case you are not able to write those, I can do it for you when I have some time.

Walti91:
Hi!

I renamed the methods to match the API documentation which is indeed missing the endpoints for categories and sections, but they do exist and are working. I also added tests for the new methods.

Kind regards

PierreBtz:
@Walti91 I opened a support ticket with Zendesk to understand why those APIs are not documented, we cannot really rely on them if the lack of documentation is on purpose and they are internal APIs for some reason. I'll keep you updated.

PierreBtz (approved):
Zendesk support confirmed there is a documentation issue on their end, so I think we can move in with this PR.
Thanks for the contribution.

PierreBtz:
Released: https://github.com/cloudbees-oss/zendesk-java-client/discussions/547

PR #552, opened 2023-03-07
Implement getUsersByExternalIds(String, String...) - resolves #548

### LINKED ISSUE #548: Enable search for users by external_id where external_id is a String (not a long)
The method to retrieve users by external ID: (`public List<User> getUsersByExternalIds(long externalId, long... externalIds)`) expects the external IDs to be `long`s, but the underlying value is a `String` in the Zendesk API 

In my use case, our external IDs are Strings, so we can't search for users by external ID using this method,..

I propose to implementing an additional `public List<User> getUsersByExternalIds(String externalId, String... externalIds)` method which 

I'll look to raise a PR for this shortly if the change would be welcomed

andy-may-at:
User external IDs are Strings in the Zendesk API, but the existing getUsersByExternalId implementation only allowed getting users by an externalId that was a long.

### DIFF
--- a/src/test/java/org/zendesk/client/v2/UserTest.java
+++ b/src/test/java/org/zendesk/client/v2/UserTest.java
@@ -1,6 +1,15 @@
 package org.zendesk.client.v2;
 
-import static com.github.tomakehurst.wiremock.client.WireMock.*;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.equalTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.equalToJson;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.getRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.ok;
+import static com.github.tomakehurst.wiremock.client.WireMock.put;
+import static com.github.tomakehurst.wiremock.client.WireMock.putRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
 import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.options;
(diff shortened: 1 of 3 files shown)


src/test/java/org/zendesk/client/v2/UserTest.java
```diff
@@ -1,20 +1,19 @@
 package org.zendesk.client.v2;
 
-import static com.github.tomakehurst.wiremock.client.WireMock.equalTo;
-import static com.github.tomakehurst.wiremock.client.WireMock.equalToJson;
-import static com.github.tomakehurst.wiremock.client.WireMock.ok;
-import static com.github.tomakehurst.wiremock.client.WireMock.put;
-import static com.github.tomakehurst.wiremock.client.WireMock.putRequestedFor;
-import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
-import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
```
PierreBtz, line 3: Nit: I'd rather see the explicit import kept (we strongly discourage using wildcard imports even in tests, see https://github.com/cloudbees-oss/zendesk-java-client/blob/master/CONTRIBUTING.adoc#imports).

(sorry I just noticed the project doesn't have a working editorconfig file to help setting up your IDE, I'll try to add it at some point).

PierreBtz (dismissed):
Lgtm, I'd like if possible to see the wildcard import removed.

src/test/java/org/zendesk/client/v2/UserTest.java
andy-may-at: Good spot - I'll get that corrected

andy-may-at: @PierreBtz - I'll be creating a local .editorconfig for this repo (to make sure that I conform to the contributing guidelines when letting IntelliJ do it's thing).
If you want, I could submit that in a separate PR

PierreBtz: Works for me :)
Thanks!

PierreBtz (approved):

PR #554, opened 2023-03-07
Fix and enable createOrUpdateUser test

### LINKED ISSUE #553: Fix and explaination for failing createOrUpdateUser test
**Describe the bug**
The createOrUpdateUser test is failing after attempting to update a phone number field on a user. It appears that the Zendesk api has additional rules for updating identity fields that are exposed on the user resource (i.e. email and phone). These rules include not allowing a duplicate value for an email field and not replacing but adding phone numbers when updated.

**To Reproduce**
Run the createOrUpdateUser test. The user modified ends up with two phone numbers instead of having the single phone number updated. The user object returned only displays the first phone number set making it appear as if the updated did not work.

**Expected behavior**
The field updated should be reflected on the object returned from the update api call. This expected behavior occurs when updating most fields. Recommended change is to update a non-identity and test for that change in the smoke test.

**Additional context**
This behavior was discovered while attempting to synchronize a large set of users with Zendesk. There are unexpected (and undocumented) behaviors with the Zendesk User endpoint including the ability to create an end-user without an email or phone number but not being able to clear both email and phone after they have been set.

PR to be submitted

colbya:
This fixes #553 and enables the createOrUpdateUser test switching the updated field from phone to details.

### DIFF
--- a/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
+++ b/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
@@ -1882,8 +1882,9 @@ public class RealSmokeTest {
     }
 
     @Test
-    @Ignore("Failing and I don't know why - caching issue ?")
-    // TODO: Fix this test
+    // Zendesk api behavior note - When update is used to change a phone number, a second phone number
+    // is added to the record. An update works as expected for non-identify fields, additional
+    // constraints exist for identify field updates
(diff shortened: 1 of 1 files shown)


PierreBtz (approved):

PierreBtz (approved):
lgtm, thanks for fixing this!

Issue #555, opened 2023-03-07
Switch to cursor based pagination for Users


colbya:
**Is your feature request related to a problem? Please describe.**
Zendesk is encouraging usage of the newer cursor based pagination especially for large result sets. We recently ran into issues with the older offset based pagination with frequent changes to our user set. The offset based pagination bases the paging off of a cached total count of users. In our case that count was being cached for 24 hours which prevent seeing additional new pages added to the user collection. A similar problem exists if a large number of users are deleted, the paging will continue to instruct you to page through empty pages until the paging matches the cached user count.

**Describe the solution you'd like**
Offset pagination is the default and cursor based pagination can be enabled by adding a `?page[size]=100` parameter to the request. The returned paging information is similar but located in a different JSON object. Support for both offset and cursor based pagination can be added by looking for the cursor based pagination first and then falling through to the existing offset based pagination.

Cursor based pagination can then be incrementally added to specific calls in this library as needed.

**Describe alternatives you've considered**
Zendesk was contacted about documenting or fixing the page count that is cached for ~24 hours and they recommended switching to cursor based pagination.

Alternatively, the paging navigation links could be ignored and paging offsets continue, even if a next link is null, until an empty page is fetched. In the case of the cached count affecting offset paging, the non-navigable pages do exist even though they never appear in the paging navigation.

PR to be submitted with a change to just the getUsers() endpoint, other endpoints could benefit from this change but are not covered by the PR

**Additional Reference**

https://developer.zendesk.com/api-reference/introduction/pagination/

https://developer.zendesk.com/documentation/api-basics/pagination/paginating-through-lists-using-cursor-pagination/

Cursor pagination includes a links and meta section in the response with an opaque cursor value that is be included to fetch the next page.

```
"meta": {
    "has_more": true,
    "after_cursor": "xxx",
    "before_cursor": "yyy"
},
"links": {
    "next": "https://example.zendesk.com/api/v2/tickets.json?page[size]=100&page[after]=xxx",
    "prev": "https://example.zendesk.com/api/v2/tickets.json?page[size]=100&page[before]=yyy"
}
```

https://developer.zendesk.com/documentation/api-basics/pagination/paginating-through-lists-using-offset-pagination/

Offset pagination includes a next_page link in the top level of the JSON response with a integer page value that returns values that start at an offset multiple from the beginning of the collection.

```
"tickets": [ ... ],
[...]

PR #556, opened 2023-03-07
Cursor based pagination added for getUsers

colbya:
Fix for #555, see issue for further details and discussion


PierreBtz (approved):
Did some testing locally, it looks like it behaves properly.
Lgtm, thanks!

PR #558, opened 2023-03-08
Add support for `Article.content_tag_ids field` - resolves #557

### LINKED ISSUE #557: Support getting/setting of Help Center article's `content_tag_ids` field
Help Center articles have a (mutable & optional) 'content_tag_ids' field
The [documentation](https://developer.zendesk.com/api-reference/help_center/help-center-api/articles/) says that the field contains '_The list of content tags attached to the article_'

This field is not currently exposed in the Article model & so cannot be read/written
Feature request is to add it to the model.

**Additional context**
N.B. a potential risk of adding a new field to the model seems to be that any existing code using the client that was doing a overwrite of an existing article (by calling `Zendesk.updateArticle(article)` (without having recently read the article in question from the API, or where the Article instance has been created by copying a known set of fields from another Article instance) would have been preserving any pre-existing content_tags_id value on the article.
But if we add the field to the model, then any code like this will now be updating the content_tags_id field to be empty
This edge-case change in behaviour seems worthwhile, but may be worth documenting in release notes.

I'm building a PR for this issue & will submit it shortly

andy-may-at:
N.B. a potential risk of adding a new field to the model seems to be that any existing code using the client that was doing a overwrite of an existing article (by calling Zendesk.updateArticle(article) (without having recently read the article in question from the API, or where the Article instance has been created by copying a known set of fields from another Article instance) would have been preserving any pre-existing content_tags_id value on the article.
But if we add the field to the model, then any code like this will now be updating the content_tags_id field to be empty
This edge-case change in behaviour seems worthwhile, but may be worth documenting in release notes.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/hc/Article.java
+++ b/src/main/java/org/zendesk/client/v2/model/hc/Article.java
@@ -40,6 +40,10 @@ public class Article implements SearchResultEntity {
     @JsonProperty("comments_disabled")
     private Boolean commentsDisabled;
 
+    /** The list of content tags attached to the article */
+    @JsonProperty("content_tag_ids")
+    private List<Long> contentTagIds;
+
     /** Whether the source (default) translation of the article is out of date */
     private Boolean outdated;
 
@@ -166,6 +170,14 @@ public class Article implements SearchResultEntity {
         this.commentsDisabled = commentsDisabled;
     }
 
+    public List<Long> getContentTagIds() {
+        return contentTagIds;
+    }
+
+    public void setContentTagIds(List<Long> contentTagIds) {
+        this.contentTagIds = contentTagIds;
+    }
+
     public Boolean getOutdated() {
         return outdated;
     }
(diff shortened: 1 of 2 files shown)


andy-may-at:
I would have been tempted to add a test for this in the RealSmokeTest, but I don't know what data's in that instance, so can't do so

andy-may-at:
D'Oh!
I've only just realised that the client doesn't support fetching of the Content Tag API resource
https://developer.zendesk.com/api-reference/help_center/help-center-api/content_tags/

I'd be happy to implement that in a PR as well.
Would you prefer me to roll the two together into a single issue/PR or to keep them separate?

PierreBtz (approved):
>  Would you prefer me to roll the two together into a single issue/PR or to keep them separate?

On one hand both changes are orthogonal so two PRs make sense, but on the other hand since your PR is the first using the content tag API, it also make sense grouping them. So I guess up to you, use the solution that requires the less work for you :)

> This edge-case change in behaviour seems worthwhile, but may be worth documenting in release notes.

Agreed, let's not over-engineer this, I'll mark this PR as breaking to remember about it in the release notes.

andy-may-at:
> On one hand both changes are orthogonal so two PRs make sense, but on the other hand since your PR is the first using the content tag API, it also make sense grouping them. So I guess up to you, use the solution that requires the less work for you

Theres some value in this change on it's own, and it's marginally simpler for me if we just merge this & I just rebase my draft PR for the rest of the content tag stuff, so I vote for that :)

PR #565, opened 2023-03-14
Users by role cursor pagination

colbya:
Added cursor pagination to getUserByRole and added a basic unit test for the method. I avoided adding admin users for the unit test. I didn't think that would be a good idea or allowed in the test environment.


PierreBtz (approved):
lgtm!

PR #582, opened 2023-04-19
Bugfix - Article.contentTagIds field now List<String> - resolves #581

### LINKED ISSUE #581: The (new in 0.22.0) articles.contentTagIds field has the wrong Type (is List<Long> - should be List<String>
**Describe the bug**
When I added the field Articles.contentTagIds in [this commit](https://github.com/cloudbees-oss/zendesk-java-client/commit/e3c895f9e03c21eb6a21e40601ebd7d49c6dd664), I trusted the example responses in the [Zendesk documentation ](https://developer.zendesk.com/api-reference/help_center/help-center-api/articles/#example-responses-1) & so added the field as a `List<Long>`

But it turns out that the articles API docs were wrong & the IDs are really `String` values (not `Long`)

### Evidence:
The [docs](https://developer.zendesk.com/api-reference/help_center/help-center-api/content_tags/#json-format) for the content_tags resource show it as a String.

And an example response from my Zendesk instance's Articles API has String values:
```
     "content_tag_ids": [
         "01GMAGH6KPDHEY79MN3S4H8V7D"
     ],
```

I would expect this bug to cause deserialisation errors when parsing article responses that contain content_tag_ids value

I'll get a PR to fix in ASAP.

andy-may-at:
The Article.contentTagIds field had been added with the incorrect type of List<Long> (because of misleading Zendesk documentation), but the IDs are actually `String`s & so can't be deserialised to Longs

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/hc/Article.java
+++ b/src/main/java/org/zendesk/client/v2/model/hc/Article.java
@@ -42,7 +42,7 @@ public class Article implements SearchResultEntity {
 
     /** The list of content tags attached to the article */
     @JsonProperty("content_tag_ids")
-    private List<Long> contentTagIds;
+    private List<String> contentTagIds;
 
     /** Whether the source (default) translation of the article is out of date */
     private Boolean outdated;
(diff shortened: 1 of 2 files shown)


duemir (approved):

johnou (approved):
lgtm matches documentation here https://developer.zendesk.com/api-reference/help_center/help-center-api/articles/

PierreBtz (approved):
Thanks for the fix, I'll put a comment in the release note since it's technically a breaking change, even if it doesn't really matter since the previous code couldn't work.

PR #594, opened 2023-06-22
Add GET timezones endpoint, add ianaTimeZone field to User

nkharabaruk:
- Added a **GET** time_zones endpoint
- Add the `ianaTimeZone` field to the **User**


src/main/java/org/zendesk/client/v2/model/TimeZone.java
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * @author nkharabaruk
+ * @since 22/06/2023 22:00
+ */
+public class TimeZone {
+
+    private String translatedName;
+    private String name;
+    private String ianaName;
+    private String url;
```
PierreBtz, line 16: I don't see this field in the documentation, but maybe I missed it, it's a bit of a pain to navigate... Do you have a link I can check?

PierreBtz (changes requested):
Thanks a lot for the contribution, I'd just like to double check the documentation of the `url` field before moving this further.

src/main/java/org/zendesk/client/v2/model/TimeZone.java
nkharabaruk: The API call returns timezones with the`url` field. 
I also don't see it in the documentation: https://developer.zendesk.com/api-reference/apps/apps-support-api/all_locations/#time-zone-object
✔️ removed

PierreBtz (approved):
lgtm, thanks for your contribution!

Issue #573, opened 2023-03-25
Cannot marshal Ticket to JSON


afm497:
**Describe the bug**
Cannot, at least using both GSON and Jackson, marshal a Ticket to JSON. Multiple fields named  **priority**. One in the Ticket itself, another in it's superclass, Request. The **priority** in Ticket is private, in Request it's protected. 

**To Reproduce**
1. Using the API, grab a Ticket object.
2. Try to marshal to JSON using either GSON or Jackson. Though it probably would apply to others, only have used those.
4. See IllegalArgumentException referencing multiple fields named "**priority**".

**Expected behavior**
Ability to marshal Ticket object to JSON for proper transmission.

**Additional context**
Not sure anymore is needed, but happy provide anything additional.

PierreBtz:
@afm497 while the shadowed field is indeed less than ideal, I'm unsure why it wouldn't work. Following code is working at producing a json:

```
    public static void main(String[] args) throws IOException {
        var ticket = new Ticket();
        ticket.setPriority(Priority.HIGH);
        var mapper = new ObjectMapper();
        mapper.writeValue(System.out, ticket);
    }
```

Please provide a fully self contained reproducer.

afm497:
I guess we differ on what "less than ideal" covers :) Just for the record, I use your library and find it very worthwhile, valuable and sincerely thank you for your efforts in creating and sharing.

Quick snippet:
`
    public static void main(String[] args) {
        new Gson().toJson(new Ticket());
    }
`
will result in
` java.lang.IllegalArgumentException: class org.zendesk.client.v2.model.Ticket declares multiple JSON fields named priority `

So while yes, one could use a custom exclusion strategy, detect this kind of thing and work around it, it seems like either the field should be fixed to only declare once if they represent the same thing or transition to distinct fields.

I'll add, now that I see the package name there, I believe this is part of your library , but if not and it's coming from elsewhere, let me know.

afm497:
Totally understand. I'll submit a PR for you here. Didn't want to do it out of the blue so glad we could connect here. Thanks very much for being open here.

afm497:
Had totally forgotten to create PR but got a notification that the issue was going stale so finally made the PR. Not sure if there is anything else I would need to do to ensure it no longer is treated as stale, but hopefully the activity addresses it.

PR #595, opened 2023-06-25
Removed variables with duplicate definitions in Ticket/Request

afm497:
Fixes cloudbees-oss/zendesk-java-client#573

Very simple/small fix, but will allow library to work correctly in an environment where GSON is used.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Ticket.java
+++ b/src/main/java/org/zendesk/client/v2/model/Ticket.java
@@ -16,8 +16,6 @@ public class Ticket extends Request implements SearchResultEntity {
     private static final long serialVersionUID = 1L;
 
     private String externalId;
-    private Type type;
-    private Priority priority;
     private String recipient;
     private Long submitterId;
     private Long assigneeId;
@@ -29,7 +27,6 @@ public class Ticket extends Request implements SearchResultEntity {
     private boolean hasIncidents;
     private Date dueAt;
     private List<String> tags;
-    private List<CustomFieldValue> customFields;
     private SatisfactionRating satisfactionRating;
     private List<Long> sharingAgreementIds;
     private List<Long> followupIds;
@@ -77,15 +74,6 @@ public class Ticket extends Request implements SearchResultEntity {
 
     public void setCollaborators(List<Collaborator> collaborators) {
         this.collaborators = collaborators;
-    }
-
-    @JsonProperty("custom_fields")
-    public List<CustomFieldValue> getCustomFields() {
-        return customFields;
-    }
-
-    public void setCustomFields(List<CustomFieldValue> customFields) {
-        this.customFields = customFields;
     }
 
     @JsonProperty("due_at")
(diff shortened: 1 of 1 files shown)


src/main/java/org/zendesk/client/v2/model/Ticket.java
```diff
@@ -16,8 +16,6 @@ public class Ticket extends Request implements SearchResultEntity {
     private static final long serialVersionUID = 1L;
```
PierreBtz, line 16: Did you test the serialization/deserialization ?
I'm not sure how Java exactly handles variable hiding with serialization, but I have a high doubt this change keeps the compatibility.

PierreBtz: I did some testing and I don't think the change will break the deserilization but it IS a breaking change:

```
    static class A implements Serializable {
        private static final long serialVersionUID = 1L;
        protected String field;
    }

    static class B extends A implements Serializable {
        private static final long serialVersionUID = 1L;
        // to remove onced we serialized
        private String field;
    }

    public static void main(String[] args) throws IOException, ClassNotFoundException {
        if(args != null && args.length >0) {
            var serializedValue = Base64.getDecoder().decode(args[0]);
            try(var bais = new ByteArrayInputStream(serializedValue)) {
                try(var ois = new ObjectInputStream(bais)) {
                    B b = (B) ois.readObject();
                    System.out.println(b);
                    System.out.println(b.field);
                }
            }
            return;
        }

        var b = new B();
        b.field = "toto";

        try (var baos = new ByteArrayOutputStream()) {
            try (var oos = new ObjectOutputStream(baos)) {
                oos.writeObject(b);
                System.out.println(Base64.getEncoder().encodeToString(baos.toByteArray()));
            }
        }
    }

}
```

Running this once to get a serialized version of `B`, then removing the private hiding field (like this PR does), I can confirm the deserialization process still works but the value of `field` is now null.

PierreBtz:
```suggestion
    private static final long serialVersionUID = -7559199410302237012L;
```

afm497:
Hey, sorry I totally missed this comment. I did test this, but I did only test it in the context of JSON marshalling. It's presumable that this could break backwards compat with binary serialization, but I do not believe there is any way around that. The error in the class is there, and there is no other way to resolve that I am aware of, at least that is relatively simple. I think the bottom line is that a bug has existed for a while here (redundant definitions of the same class member with differing accessibility. One is private, one is protected and they are exposing the same piece of data.

PierreBtz:
I agree with you, but without my suggestion of changing the `serialVersionUID`, a code de-serializing a Ticket serialized with an older version would silently work, giving undefined behavior. I'd rather see such code fail, therefore we should change the `serialVersionUID`.

afm497:
Ahhh, definitely I agree wholeheartedly. Didn't see that suggestion, apologies. I'll update accordingly. 

Thanks,
Ray

PierreBtz (approved):
great, thanks a lot for your contribution!

PR #597, opened 2023-06-26
fix: Remove length constraint from organization external id lookup

enismustafaj:


PierreBtz (changes requested):
I don't see any trace in the Zendesk documentation, nor in this repository history, explaining why this limitation was added.

I suppose it's therefore ok to lift it, but I'd like to have a test cover this use case.

src/test/java/org/zendesk/client/v2/RealSmokeTest.java
```diff
@@ -1583,6 +1584,21 @@ public void createOrganizationMemberships() throws Exception {
...
+    public void lookupOrganizationByExternalId() throws Exception {
+        createClientWithTokenOrPassword();
+
+        Organization newOrganization = newTestOrganization();
+        newOrganization.setExternalId("i");
+        Organization resultOrganization = instance.createOrganization(newOrganization);
+        assertNotNull(resultOrganization);
+
+        Iterable<Organization> or =  instance.lookupOrganizationsByExternalId("i");
+        assertTrue(or.iterator().hasNext());
```
PierreBtz, line 1597: Only makes sure something is returned, we should also make sure the size is ok, what about the following?

```suggestion
        assertEquals(1, StreamSupport.stream(or.spliterator(), false).count());
```

PierreBtz (commented):
Change looks good, I'd like for the test to be more precise in its assertion. Proposed something, please tell me what you think about it.

src/test/java/org/zendesk/client/v2/RealSmokeTest.java
enismustafaj: Added the change in the test.

PierreBtz (approved):
alright, perfect, thanks!

PR #608, opened 2023-08-16
Sircpl/cbp by default

cryptomail:
Addresses https://github.com/cloudbees-oss/zendesk-java-client/issues/588

### DIFF
--- a/src/main/java/org/zendesk/client/v2/DefaultUserAgent.java
+++ b/src/main/java/org/zendesk/client/v2/DefaultUserAgent.java
@@ -1,0 +1,39 @@
+package org.zendesk.client.v2;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+import java.util.regex.Pattern;
+
+public class DefaultUserAgent {
+  private static final Properties ZDJ_PROPERTIES = loadProperties();
+  private static final Pattern VERSION_PATTERN = Pattern.compile("[\\w-.]+");
+  private final String userAgent;
+
+  private static Properties loadProperties() {
+    try (InputStream is =
+        DefaultUserAgent.class.getResourceAsStream("/zendesk-java-client.properties")) {
+      Properties properties = new Properties();
+      properties.load(is);
+      return properties;
+    } catch (IOException ioe) {
+      throw new RuntimeException(ioe);
+    }
+  }
+
+  public DefaultUserAgent() {
+    this(ZDJ_PROPERTIES.getProperty("version"));
+  }
+
+  public DefaultUserAgent(String version) {
+    StringBuilder sb = new StringBuilder("zendesk-java-client");
+    if (version != null && VERSION_PATTERN.matcher(version).matches()) {
+      sb.append('/').append(version);
+    }
+    this.userAgent = sb.toString();
+  }
+
+  public String toString() {
+    return this.userAgent;
+  }
+}
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -116,6 +116,7 @@ import org.zendesk.client.v2.model.targets.UrlTarget;
  */
 public class Zendesk implements Closeable {
   private static final String JSON = "application/json; charset=UTF-8";
+  private static final String USER_AGENT_HEADER = "User-Agent";
   private static final DefaultAsyncHttpClientConfig DEFAULT_ASYNC_HTTP_CLIENT_CONFIG =
       new DefaultAsyncHttpClientConfig.Builder().setFollowRedirect(true).build();
   private final boolean closeClient;
@@ -187,6 +188,7 @@ public class Zendesk implements Closeable {
       }
       this.realm = null;
     }
+    headers.putIfAbsent(USER_AGENT_HEADER, new DefaultUserAgent().toString());
     this.headers = Collections.unmodifiableMap(headers);
     this.cbpPageSize = cbpPageSize;
     this.mapper = createMapper();
@@ -210,6 +212,7 @@ public class Zendesk implements Closeable {
       throw new IllegalStateException(
           "Cannot specify token or password without specifying username");
     }
+    headers.putIfAbsent(USER_AGENT_HEADER, new DefaultUserAgent().toString());
     this.headers = Collections.unmodifiableMap(headers);
     this.cbpPageSize = cbpPageSize;
     this.mapper = createMapper();
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -1,5 +1,8 @@
 package org.zendesk.client.v2;
 
+import static java.util.Collections.singletonList;
+import static org.asynchttpclient.Param.map2ParamList;
+
 import com.damnhandy.uri.template.UriTemplate;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonProcessingException;
@@ -29,7 +32,6 @@ import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
-
 import org.asynchttpclient.AsyncCompletionHandler;
 import org.asynchttpclient.AsyncHttpClient;
 import org.asynchttpclient.DefaultAsyncHttpClient;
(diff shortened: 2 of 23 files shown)


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -3783,6 +3929,27 @@ private TemplateUri tmpl(String template) {
...
+  private Uri cbp(String path, Param... queryParameters) {
+    return cbp(path, Arrays.asList(queryParameters));
+  }
+
+  private Uri cbp(String path, Iterable<Param> queryParameters) {
+    Objects.requireNonNull(path, "Path cannot be null");
+    if (path.indexOf('?') != -1) {
+      throw new IllegalArgumentException("Path cannot contain a query string");
+    }
+    // No need to encode parameters. Handled by AHC.
```
cryptomail, line 3941: This did catch my eye but the comments are awesome.  I believe 👽 🛸

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -4167,6 +4334,7 @@ public boolean hasNext() {
           if (nextPage == null || nextPage.equalsIgnoreCase("null")) {
             return false;
           }
+          System.out.println(nextPage); // FIXME DO NOT COMMIT
```
cryptomail, line 4337: Please remove this line

cryptomail (commented):
LGTM let's put it up for review to the larger group.
Please burn that one log tho.... 😆 🔥 🪵

src/main/java/org/zendesk/client/v2/DefaultUserAgent.java
```diff
@@ -0,0 +1,39 @@
+package org.zendesk.client.v2;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+import java.util.regex.Pattern;
+
+public class DefaultUserAgent {
```
PierreBtz, line 8: I'm not against this change, but it's unrelated to the migration to CBP.
This PR is already big enough, I'd rather see this in a different one.
Two notes (that we discuss later in a the split PR):
* end user can easily change this so you're not guaranteed to get the information server side.
* AFAIR the jar manifest should already contain the artifact version, it might be simpler to pick it up this way.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -664,6 +710,15 @@ public Metric getTicketMetric(long id) {
             handle(Metric.class, "ticket_metric")));
   }
 
+  public Iterable<Audit> getTicketAudits(Map<String, List<String>> parameters) {
```
PierreBtz, line 713: I'm under the impression this adds the ability to push additional parameters to the endpoint. 
While I understand the idea (not wait for the library to be updated if a parameter is added to the library, I'm not sure this is desirable, since it goes against what was done up until now in the library. At the very least this should be split into its separate PR for further discussion.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -782,9 +846,31 @@ public void deleteTicketField(long id) {
...
+        handleList(SuspendedTicket.class, "suspended_tickets"));
+  }
+
   public Iterable<SuspendedTicket> getSuspendedTickets() {
     return new PagedIterable<>(
-        cnst("/suspended_tickets.json"), handleList(SuspendedTicket.class, "suspended_tickets"));
+        cbp("/suspended_tickets.json"), handleList(SuspendedTicket.class, "suspended_tickets"));
+  }
+
+  public Iterable<Activity> getActivities(Map<String, List<String>> parameters) {
```
PierreBtz, line 860: Looks like a new endpoint support, good idea, but could we move it into its own PR?

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -782,9 +846,31 @@ public void deleteTicketField(long id) {
...
+    return new PagedIterable<>(
+        cbp("/activities.json", map2ParamList(parameters)),
+        handleList(Activity.class, "activities"));
+  }
+
+  public Iterable<Activity> getActivities() {
+    return new PagedIterable<>(cbp("/activities.json"), handleList(Activity.class, "activities"));
+  }
+
+  public Iterable<RecipientAddress> getRecipientAddresses() {
```
PierreBtz, line 870: Ditto

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -938,24 +1024,33 @@ public void deleteTarget(long targetId) {
...
   public Iterable<Trigger> getTriggers(
       String categoryId, boolean active, String sortBy, SortOrder sortOrder) {
-    return new PagedIterable<>(
-        tmpl("/triggers.json{?category_id,active,sort_by,sort_order}")
-            .set("category_id", categoryId)
-            .set("active", active)
-            .set("sort_by", sortBy)
-            .set("sort_order", sortOrder.getQueryParameter()),
-        handleList(Trigger.class, "triggers"));
+    Map<String, List<String>> parameters = new HashMap<>();
```
PierreBtz, line 1038: Another reason why I'm not certain we should allow to pass any parameter, it adds additional logic to the existing code.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -3783,6 +3965,27 @@ private TemplateUri tmpl(String template) {
...
+  private Uri cbp(String path, Iterable<Param> queryParameters) {
+    Objects.requireNonNull(path, "Path cannot be null");
+    if (path.indexOf('?') != -1) {
+      throw new IllegalArgumentException("Path cannot contain a query string");
+    }
+    // No need to encode parameters. Handled by AHC.
+    // See
+    // https://github.com/AsyncHttpClient/async-http-client/blob/async-http-client-project-2.12.3/client/src/main/java/org/asynchttpclient/RequestBuilderBase.java#L598-L609
+    StringBuilder sb =
+        new StringBuilder(url).append(path).append("?page[size]=").append(cbpPageSize);
```
PierreBtz, line 3981: This project relies on https://github.com/damnhandy/Handy-URI-Templates to build URIs. It removes the hassle of manually building the string (which is error prone too). See the README in the link I provided, it contains useful examples.

PierreBtz (changes requested):
Thanks for the tremendous effort put into this PR.

Our first goal to get this merged would be to extract the various, unrelated developments, into their own PRs:

* custom user agent
* random query parameter support for various endpoints -> as explained in my inline comments, I have mixed feelings about this one, I'm unsure if this is desirable yet.
* Activity endpoint support
* RecipientAddress endpoint support
* Tags endpoint support

Back to the CBP changes themselves:
* I'm ok to go with a breaking release, since the alternative to have a detail of implementation in the API method names is not desirable in my opinion.
* I'd rather not start manually building URIs when we have an URI templating engine available

Again, thanks a lot for the time and effort you put into this!

aheritier:
What an awesome contribution ! thanks a lot @sircpl 
I agree with @PierreBtz that it's difficult to read between CBP changes, additional APIs, spotless fixes...
@PierreBtz can you verify if the tests are passing (I don't have the permission to trigger them).
This API is still in 0.x for many bad reasons thus breaking again the compatibility is ok but maybe we should really consider to move to 1.x after these improvements

src/main/java/org/zendesk/client/v2/Zendesk.java
sircpl: The reason why the uri templates were not used is that they do not support `page[size]` as a parameter name:
```
  UriTemplate invalid = UriTemplate.fromTemplate("https://yourdomain.zendesk.com/api/v2/groups?{page[size],foo}");
  invalid.set("page[size]", 2);
  invalid.set("foo", "bar");
  System.out.println(invalid.expand());

```
produces `com.damnhandy.uri.template.MalformedUriTemplateException: The variable name page[size] contains invalid characters`

I believe this is the same reason this [code exists](https://github.com/cloudbees-oss/zendesk-java-client/blob/master/src/main/java/org/zendesk/client/v2/Zendesk.java#L1074-L1076).

It may not be an issue though once I get rid of the ability to pass query parameters.

sircpl:
I will split the PR up as requested

sircpl:
I've split this PR into three separate PRs:

- Spotless fixups https://github.com/cloudbees-oss/zendesk-java-client/pull/612 (not related to CBP changes)
- CBP support https://github.com/cloudbees-oss/zendesk-java-client/pull/614
- https://github.com/cloudbees-oss/zendesk-java-client/pull/613

I will take care of adding the newly supported endpoints in this PR after the above are merged (assuming they're okay 😸)

aheritier:
Awesome @sircpl 
Really appreciated

PierreBtz:
Supersed by #612 #613 #614

PR #612, opened 2023-08-30
Spotless fixes

sircpl:
These are the result of running `mvn spotless:apply` as suggested by the plugin


aheritier (approved):
+1 as far as it is coherent and  automated

PierreBtz:
Build is not happy, it seems the formatting was correct.
I think I understood what went on here. When you started to work on CBP you created a branch out of the version of master using a different version of the spotless plugin which was using a slightly different formatting.

Since then I merged some changes (https://github.com/cloudbees-oss/zendesk-java-client/commit/6a5536f91673d4631331178c287f2a6c3f8f32b4).

PR #613, opened 2023-08-30
Add default user agent header

sircpl:
Adds a default user-agent header to requests made by the client. This helps identify traffic coming from the client, but doesn't prevent users from supplying their own user-agent header.

This is slightly modified from the version found in https://github.com/cloudbees-oss/zendesk-java-client/pull/608. It requires that the version be written into the manifest which is accomplished by `addDefaultImplementationEntries`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -256,6 +256,9 @@
         <version>3.3.0</version>
         <configuration>
           <archive>
+            <manifest>
+              <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+            </manifest>
             <manifestEntries>
               <Automatic-Module-Name>org.zendesk.client.v2</Automatic-Module-Name>
             </manifestEntries>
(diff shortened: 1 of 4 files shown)


aheritier (approved):
If it's useful on your side to better identify the different client API and versions let's go

duemir (commented):
I also agree it is useful. However, I am not sure about the implementation. Instantiating an object to build a sting seems like an overkill to me. I probably would have added package class `Headers` with `USER_AGENT` header constant and static method `defaultUserAgent`. WDYT @PierreBtz ?

PierreBtz (approved):
@duemir There is additional code needed to retrieve the artifact version. I suggested an alternate approach (https://github.com/cloudbees-oss/zendesk-java-client/pull/608#discussion_r1309887465) reading from the MANIFEST (that should be filled by maven at build time IIRC), but it's marginally simpler.

It works as is (tested), so I'm happy, but if you have in mind another simpler approach, I'm interested.

duemir:
@PierreBtz It is OK. I'll merge, then.

PR #614, opened 2023-08-30
Use cursor based pagination for endpoints that support it

sircpl:
Use CBP for all endpoints that support it. 

This PR does not include any of the newly added endpoints in https://github.com/cloudbees-oss/zendesk-java-client/pull/608. New endpoints supporting CBP will be added _after_ this PR is merged (since this PR contains the `cbp` method).

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -2394,9 +2394,7 @@ public class Zendesk implements Closeable {
     return getSearchResults(type, query, Collections.emptyMap());
   }
 
-  /**
-   * @deprecated Use {@link #getSearchResults(Class, String, Map)} instead.
-   */
+  /** @deprecated Use {@link #getSearchResults(Class, String, Map)} instead. */
   @Deprecated
   public <T extends SearchResultEntity> Iterable<T> getSearchResults(
       Class<T> type, String query, String params) {
@@ -2852,9 +2850,7 @@ public class Zendesk implements Closeable {
     return complete(submit(req("GET", cnst("/help_center/locales.json")), handle(Locales.class)));
   }
 
-  /**
-   * @deprecated Use {@link Zendesk#listHelpCenterLocales()} instead
-   */
+  /** @deprecated Use {@link Zendesk#listHelpCenterLocales()} instead */
   @Deprecated
   public List<String> getHelpCenterLocales() {
     return listHelpCenterLocales().getLocales();
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -115,6 +115,7 @@ public class Zendesk implements Closeable {
   private final String url;
   private final String oauthToken;
   private final Map<String, String> headers;
+  private final int cbpPageSize;
   private final ObjectMapper mapper;
   private final Logger logger;
   private boolean closed = false;
@@ -156,7 +157,8 @@ public class Zendesk implements Closeable {
       String url,
       String username,
       String password,
-      Map<String, String> headers) {
+      Map<String, String> headers,
+      int cbpPageSize) {
     this.logger = LoggerFactory.getLogger(Zendesk.class);
     this.closeClient = client == null;
     this.oauthToken = null;
@@ -177,11 +179,16 @@ public class Zendesk implements Closeable {
       this.realm = null;
     }
     this.headers = Collections.unmodifiableMap(headers);
+    this.cbpPageSize = cbpPageSize;
     this.mapper = createMapper();
   }
 
   private Zendesk(
-      AsyncHttpClient client, String url, String oauthToken, Map<String, String> headers) {
+      AsyncHttpClient client,
+      String url,
+      String oauthToken,
+      Map<String, String> headers,
+      int cbpPageSize) {
     this.logger = LoggerFactory.getLogger(Zendesk.class);
     this.closeClient = client == null;
     this.realm = null;
@@ -195,7 +202,7 @@ public class Zendesk implements Closeable {
           "Cannot specify token or password without specifying username");
     }
     this.headers = Collections.unmodifiableMap(headers);
-
+    this.cbpPageSize = cbpPageSize;
     this.mapper = createMapper();
   }
 
@@ -246,8 +253,8 @@ public class Zendesk implements Closeable {
         handleList(clazz, "job_statuses"));
   }
 
-  public List<Brand> getBrands() {
-    return complete(submit(req("GET", cnst("/brands.json")), handleList(Brand.class, "brands")));
+  public Iterable<Brand> getBrands() {
+    return new PagedIterable<>(cbp("/brands.json"), handleList(Brand.class, "brands"));
   }
 
   public TicketForm getTicketForm(long id) {
@@ -356,11 +363,9 @@ public class Zendesk implements Closeable {
             req("GET", tmpl("/tickets/{id}.json").set("id", id)), handle(Ticket.class, "ticket")));
   }
 
-  public List<Ticket> getTicketIncidents(long id) {
-    return complete(
-        submit(
-            req("GET", tmpl("/tickets/{id}/incidents.json").set("id", id)),
-            handleList(Ticket.class, "tickets")));
+  public Iterable<Ticket> getTicketIncidents(long id) {
+    return new PagedIterable<>(
+        cbp("/tickets/{id}/incidents.json").set("id", id), handleList(Ticket.class, "tickets"));
   }
 
   public List<User> getTicketCollaborators(long id) {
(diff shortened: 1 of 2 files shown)


aheritier (approved):
LGTM but I cannot test them. As said in the
Last PR, even if we don't implement 100% of zendesk APIs it might be great to start to provide a stable API and to have a formal versioning. 

(Especially if we can get some official support from Zendesk like here)

PierreBtz:
Sorry I didn't have time to progress on this. I'll progress on the review and testing next week.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -246,8 +253,8 @@ public ListenableFuture<List<JobStatus>> getJobStatusesAsync(List<JobStatus> sta
         handleList(clazz, "job_statuses"));
   }
 
-  public List<Brand> getBrands() {
-    return complete(submit(req("GET", cnst("/brands.json")), handleList(Brand.class, "brands")));
+  public Iterable<Brand> getBrands() {
+    return new PagedIterable<>(cbp("/brands.json"), handleList(Brand.class, "brands"));
```
PierreBtz, line 257: This change doesn't match the current Rest API documentation.

According to it:

* a endpoint may or may not support CBP, I need to refer to the endpoint documentation (https://developer.zendesk.com/documentation/api-basics/pagination/paginating-through-lists-using-cursor-pagination/).
* refering to the brand endpoint, the only example I see uses OBP (https://developer.zendesk.com/api-reference/ticketing/account-configuration/brands/)

This is of importance, because according to the documentation, depending on the endpoint we need to check the endpoint doc to choose between "Paginating with the after cursor" and "Paginating with the next link".

Usually I would open a ticket with Zendesk support to get clarification and update of the doc if necessary, but in this case I trust you'll have a quicker answer by yourself :)

PierreBtz (changes requested):
The changes look good to me, but there is a big issue in that the Rest API documentation hasn't been properly updated for me to validate the changes (except by trial and error...).

The doc clearly states that CBP works in two different ways (`Paginating with the next link` and `Paginating with the after cursor`), and that we should refer to each individual endpoint documentation to know what is supported, but, except if I missed something, there are a large number of endpoints that either do not specify anything, or show examples using the old OBP.

This makes it hard to entirely review this change, and be confident things won't break in the future.

This has nothing to do with your PR per-se, thanks a lot for the work you put into this, but I think we should clarify and have the documentation updated. 
I can take care of this through the support but maybe things will go faster if you ask internally?

PierreBtz:
Thanks for the article, it indeed gives a list of endpoints supporting CBP, I'll use that as a reference for this PR, hoping to see the rest api doc updated in the future :)

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1500,18 +1505,18 @@ public Iterable<AgentRole> getCustomAgentRoles() {
 
   public Iterable<org.zendesk.client.v2.model.Request> getRequests() {
     return new PagedIterable<>(
-        cnst("/requests.json"), handleList(org.zendesk.client.v2.model.Request.class, "requests"));
+        cbp("/requests.json"), handleList(org.zendesk.client.v2.model.Request.class, "requests"));
```
PierreBtz, line 1508: Not in the list provided in [support.zendesk.com/hc/en-us/articles/5591904358938-New-limits-for-offset-based-pagination#h_01GNCT8Z7DJ7G4TQ3CFFXQ007Z](https://support.zendesk.com/hc/en-us/articles/5591904358938-New-limits-for-offset-based-pagination#h_01GNCT8Z7DJ7G4TQ3CFFXQ007Z)

The endpoint documentation doesn't include CBP either https://developer.zendesk.com/api-reference/ticketing/tickets/ticket-requests/#list-requests

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1500,18 +1505,18 @@ public Iterable<AgentRole> getCustomAgentRoles() {
...
   }
 
   public Iterable<org.zendesk.client.v2.model.Request> getOpenRequests() {
     return new PagedIterable<>(
-        cnst("/requests/open.json"),
+        cbp("/requests/open.json"),
```
PierreBtz, line 1513: ditto

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1500,18 +1505,18 @@ public Iterable<AgentRole> getCustomAgentRoles() {
...
         handleList(org.zendesk.client.v2.model.Request.class, "requests"));
   }
 
   public Iterable<org.zendesk.client.v2.model.Request> getSolvedRequests() {
     return new PagedIterable<>(
-        cnst("/requests/solved.json"),
+        cbp("/requests/solved.json"),
```
PierreBtz, line 1519: ditto

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -2320,13 +2320,8 @@ public void deleteTopic(Topic topic) {
...
 
   // https://support.zendesk.com/hc/communities/public/posts/203464106-Managing-Organization-Memberships-via-the-Zendesk-API
-  public List<OrganizationMembership> getOrganizationMembershipByUser(long user_id) {
-    return complete(
-        submit(
-            req(
-                "GET",
-                tmpl("/users/{user_id}/organization_memberships.json").set("user_id", user_id)),
-            handleList(OrganizationMembership.class, "organization_memberships")));
+  public Iterable<OrganizationMembership> getOrganizationMembershipByUser(long user_id) {
```
PierreBtz, line 2323: Nice catch, we could probably deprecate this method for later deletion, but outside of the scope of this PR.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -2583,7 +2580,7 @@ public Iterable<DynamicContentItemVariant> getDynamicContentItemVariants(
       DynamicContentItem item) {
     checkHasId(item);
     return new PagedIterable<>(
-        tmpl("/dynamic_content/items/{id}/variants.json").set("id", item.getId()),
+        cbp("/dynamic_content/items/{id}/variants.json").set("id", item.getId()),
```
PierreBtz, line 2583: Not in the list provided in [support.zendesk.com/hc/en-us/articles/5591904358938-New-limits-for-offset-based-pagination#h_01GNCT8Z7DJ7G4TQ3CFFXQ007Z](https://support.zendesk.com/hc/en-us/articles/5591904358938-New-limits-for-offset-based-pagination#h_01GNCT8Z7DJ7G4TQ3CFFXQ007Z)

The endpoint documentation doesn't include CBP either https://developer.zendesk.com/api-reference/ticketing/ticket-management/dynamic_content/#list-items

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -2537,7 +2534,7 @@ public SatisfactionRating createSatisfactionRating(
 
   public Iterable<DynamicContentItem> getDynamicContentItems() {
     return new PagedIterable<>(
-        cnst("/dynamic_content/items.json"), handleList(DynamicContentItem.class, "items"));
+        cbp("/dynamic_content/items.json"), handleList(DynamicContentItem.class, "items"));
```
PierreBtz, line 2537: Not in the list provided in [support.zendesk.com/hc/en-us/articles/5591904358938-New-limits-for-offset-based-pagination#h_01GNCT8Z7DJ7G4TQ3CFFXQ007Z](https://support.zendesk.com/hc/en-us/articles/5591904358938-New-limits-for-offset-based-pagination#h_01GNCT8Z7DJ7G4TQ3CFFXQ007Z)

The endpoint documentation doesn't include CBP either [developer.zendesk.com/api-reference/ticketing/ticket-management/dynamic_content/#list-items](https://developer.zendesk.com/api-reference/ticketing/ticket-management/dynamic_content/#list-items)

PierreBtz (changes requested):
We're nearly ready to merge.
I still found a handful of endpoints without any documentation suggesting CBP support (either in the announcement post or the endpoint doc). Once this is sorted out, I think this will be ready.

PierreBtz:
I took the liberty of refreshing your branch onto the latest master commits, and also fix the formatting, so that we have a fresh CI run.

Btw I finally understood the trouble you had with spotless. It turns out the formatting is different between a jdk8 and a jdk11/17... The project is still compatible with Java 8, so you need to format using a jdk8.

cryptomail:
@PierreBtz We're going to address your concerns.  Thanks so much for looking all this up.
You're 100% correct that article is incomplete.  Just from cursory inspection and execution of the endpoint https://mydomain.zendesk.com/api/v2/dynamic_content/items?page[size]=1 returns CBP as the output protocol.
Again you have pointed out clean deficiencies within the documentation of our API that must be addressed.

src/main/java/org/zendesk/client/v2/Zendesk.java
snatchblock: @PierreBtz The support for Cursor Model is determined based on the endpoint use case and the depth of records we saw in the system for all customers. 

We are working on your feedback and currently evaluating the pagination information in the developer docs. We will get the right details added to every endpoint. There is active work in progress now on this outcome. Should be done soon.

PierreBtz:
@cryptomail @sircpl any news on your side, you did a good and important work and I'm a bit sad to see it blocked for some documentation issues.

PierreBtz:
@cryptomail @sircpl I had a quick look and the endpoints are still not officially compatible with CBP. My proposal to get this moving is to partially revert the problematic endpoints and merge the rest of the work. Would that work for you? I can take care of the partial revert.

PierreBtz (approved):
As discussed privately with the Zendesk team, I'm going ahead with the merge in advance of the documentation update.

PR #619, opened 2023-09-12
Support List Locales operation

topero:
Hello,

We needed to support the [List Locales operation](https://developer.zendesk.com/api-reference/ticketing/account-configuration/locales/#list-locales) and this is the change that we implemented.

We tried to follow the contributing guidelines, but feel free to comment if we missed something, please.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Locale.java
+++ b/src/main/java/org/zendesk/client/v2/model/Locale.java
@@ -1,0 +1,95 @@
+package org.zendesk.client.v2.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.Serializable;
+import java.util.Date;
+
+/**
+ * See <a
+ * href="https://developer.zendesk.com/api-reference/ticketing/account-configuration/locales/">
+ * Locales in Zendesk API </a>
+ *
+ * @since FIXME
+ */
+public class Locale implements Serializable {
+
+  private static final long serialVersionUID = 1L;
+
+  private Date createdAt;
+  private Long id;
+  private String locale;
+  private String name;
+  private Date updatedAt;
+  private String url;
+
+  @JsonProperty("created_at")
+  public Date getCreatedAt() {
+    return createdAt;
+  }
+
+  public void setCreatedAt(Date createdAt) {
+    this.createdAt = createdAt;
+  }
+
+  public Long getId() {
+    return id;
+  }
+
+  public void setId(Long id) {
+    this.id = id;
+  }
+
+  public String getLocale() {
+    return locale;
+  }
+
+  public void setLocale(String locale) {
+    this.locale = locale;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  @JsonProperty("updated_at")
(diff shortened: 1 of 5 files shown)


PierreBtz (changes requested):
@topero thanks for the addition. Change lgtm, I'd like to have a test to cover any change that might happen in the API in the future.

I know it's a pain to write a test for this library since we cannot grant access to our internal ZD testing instance to external contributors, so I can take care of the test if needed.

topero:
Sorry for the late reply, @PierreBtz.
A couple of unit tests have been added to the pull request.

In addition, a few more attributes have been added to the "Locale" class because they were returned by our Zendesk instance although I haven't found them in [Zendesk documentation](https://developer.zendesk.com/api-reference/ticketing/account-configuration/locales/#json-format).

PierreBtz:
@topero sorry I had a rather busy week, I'll look at the PR beginning of next week in more details but at the first glance it looks good.
My only reservation is that you are adding undocumented fields, I usually tend to open a ticket with Zendesk support to have them confirm they really exist and update their documentation before accepting such change.
If you need the fields, it's allright, we can do it but it will delay the merge of the PR. If you don't need them, I'd suggest to not put them, this will fast-track this.

src/main/java/org/zendesk/client/v2/model/Locale.java
```diff
@@ -0,0 +1,143 @@
...
+ * Locales in Zendesk API </a>
+ *
+ * @since FIXME
+ */
+public class Locale implements Serializable {
+
+  private static final long serialVersionUID = 1L;
+
+  private Date createdAt;
+  private Boolean isDefault;
```
PierreBtz, line 19: undocumented by ZD

src/main/java/org/zendesk/client/v2/model/Locale.java
```diff
@@ -0,0 +1,143 @@
...
+  private Long id;
+  private String locale;
+  private String name;
+  private String nativeName;
```
PierreBtz, line 23: undocumented by ZD

src/main/java/org/zendesk/client/v2/model/Locale.java
```diff
@@ -0,0 +1,143 @@
...
+  private String presentationName;
```
PierreBtz, line 24: undocumented by ZD

src/main/java/org/zendesk/client/v2/model/Locale.java
```diff
@@ -0,0 +1,143 @@
...
+  private Boolean rtl;
```
PierreBtz, line 25: undocumented by ZD

PierreBtz (commented):
lgtm once we sort out the fields that are not documented by Zendesk.

topero:
I'm sorry for the very very long delay to answer you, @PierreBtz .

On 2 October, a Zendesk Technical Account Manager was reached out with regard to this topic. In my unicorn-and-rainbows world, the documentation is updated and this PR might be kept as it was. But the reality is that I've seen no answer so far. So the undocumented fields have been removed and the tests, adjusted.

Let me know if something else needs to be modified, please.

PierreBtz (approved):
Great! Thanks for the contribution!

PR #626, opened 2023-10-16
Updated subscriptions methods to use cursor pagination

jonjarvis:
I have updated the three subscription methods to use the recommended cursor pagination in the request instead of the default offset pagination (which has a 100 page limit).

https://developer.zendesk.com/api-reference/help_center/help-center-api/content_subscriptions/#list-section-subscriptions


jonjarvis:
I just realized there is already a pull request on this subject that dealt with the change more holistically. This commit did however get me over the obstacle I was facing.

PierreBtz:
@jonjarvis thanks for the contribution. As you noted, this is a subset of https://github.com/cloudbees-oss/zendesk-java-client/pull/614. I'd rather see the holistic implementation get merged. I'll try to see if we can lift the blocker there.

jonjarvis:
Closing pull request because #614 is a more holistic solution.

Issue #629, opened 2023-10-19
Support for Safe Updates


PierreBtz:
### Discussed in https://github.com/cloudbees-oss/zendesk-java-client/discussions/627

<div type='discussions-op-text'>

<sup>Originally posted by **btpnlsl** October 17, 2023</sup>
I was wondering if there were any plans for the Zendesk Client to support safe updates as described in the [Protecting against ticket update collisions](https://developer.zendesk.com/documentation/ticketing/managing-tickets/creating-and-updating-tickets/#protecting-against-ticket-update-collisions) documentation?</div>

PR #630, opened 2023-10-19
Add Ticket properties to support safe update

btpnlsl:
#629 
This change adds the `safe_update` and `updated_stamp` properties to the `Ticket` class to support Zendesk's protection against ticket update collisions as described in [the Zendesk documentation](https://developer.zendesk.com/documentation/ticketing/managing-tickets/creating-and-updating-tickets/#protecting-against-ticket-update-collisions).

These properties will only serialize into JSON when `safe_update` is set to TRUE.

I've added some unit tests for serialization and have performed some manual tests to validate:
- Attempting to update a ticket with `safe_update` set to true can succeed.
- Attempting to update an out-of-date ticket with `safe_update` set to true will result in a HTTP CONFLICT status.
- Attempting to update an out-of-date ticket with `safe_update` set to false will succeed.

### DIFF
--- a/src/main/java/org/zendesk/client/v2/model/Ticket.java
+++ b/src/main/java/org/zendesk/client/v2/model/Ticket.java
@@ -222,7 +222,7 @@ public class Ticket extends Request implements SearchResultEntity {
   }
 
   /**
-   * The safe_update & update_stamp parameters are used by the Zendesk API to protect against update
+   * The safe_update &amp; update_stamp parameters are used by the Zendesk API to protect against update
    * collisions. If the safe_update parameter is set to true and Zendesk detects that a ticket has
    * been updated since the time specified in update_stamp then it will return an HTTP CONFLICT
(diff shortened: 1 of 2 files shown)


src/main/java/org/zendesk/client/v2/model/Ticket.java
```diff
@@ -218,6 +221,34 @@ public void setIsPublic(Boolean isPublic) {
     this.isPublic = isPublic;
   }
 
+  /**
+   * The safe_update & update_stamp parameters are used by the Zendesk API to protect against update
```
PierreBtz, line 225: I believe the javadoc generation error is due to the unencoded &. Let's try:

```suggestion
   * The safe_update &amp; update_stamp parameters are used by the Zendesk API to protect against update
```

PierreBtz (approved):
lgtm, thanks for the contribution. I'll merge once CI is green.

PierreBtz (approved):

PR #632, opened 2023-11-06
[bug] Revert a test that highlights a bug in the Zendesk Rest API

PierreBtz:
This additional assertion was (righfully) added in https://github.com/cloudbees-oss/zendesk-java-client/pull/614. However the sort_by parameter is incorrect, it should be alphabetical as per https://developer.zendesk.com/api-reference/ticketing/business-rules/triggers/#list-triggers. However doing so (in CPB mode only), results in a Zendesk error:

```
{
    "error": {
        "title": "DatabaseError",
        "message": "Unknown database error"
    }
}
```

This was missing by the CI pipeline (to be investigated too). This is also a blocker for the release.

### DIFF
--- a/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
+++ b/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
@@ -314,12 +314,8 @@ public class RealSmokeTest {
   public void getTriggersWithParameters() throws Exception {
     createClientWithTokenOrPassword(2);
     int count = 0;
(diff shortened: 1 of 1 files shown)


cryptomail:
Thanks, working internally with the team to see how to best proceed. Thanks again for your attention on this.

PierreBtz:
Superseded by https://github.com/cloudbees-oss/zendesk-java-client/pull/634#event-10878757410

PR #633, opened 2023-11-06
Fix sorting when using cbp for Triggers; Fix Trigger model to match documented JSON format

sircpl:
Smoke test was previously providing a `sort_by` of `title` which is not a supported sorting attribute (for cursor based pagination, or legacy pagination). My triggers that I used to test just happened to be sorted by title allowing the test to pass despite the invalid sort parameter 🤦 

This brought to light the fact that the correct parameter for CBP is `sort`. The documentation is in the process of being updated, and this PR will bring the implementation in line with the documentation. I modified the test to cover sort order in both directions (asc, dec)

I also fixed an issue with the `Trigger` model where several fields were not mapped properly. They now match the documented json format https://developer.zendesk.com/api-reference/ticketing/business-rules/triggers/#json-format


sircpl:
Closing as I've discovered a bug in the cbp implementation in the endpoint. Will put up a pr to revert triggers cbp portion of https://github.com/cloudbees-oss/zendesk-java-client/pull/614

PR #634, opened 2023-11-06
Revert changes to getTriggersWithParameters due to CBP bug in API

sircpl:
Until we can address the CBP bug for triggers let's revert this portion of https://github.com/cloudbees-oss/zendesk-java-client/pull/614

### DIFF
--- a/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
+++ b/src/test/java/org/zendesk/client/v2/RealSmokeTest.java
@@ -317,7 +317,7 @@ public class RealSmokeTest {
     String title = null;
     for (Trigger t : instance.getTriggers(null, true, "alphabetical", SortOrder.ASCENDING)) {
       if (title != null) {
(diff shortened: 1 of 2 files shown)


src/test/java/org/zendesk/client/v2/RealSmokeTest.java
```diff
@@ -315,9 +315,9 @@ public void getTriggersWithParameters() throws Exception {
     createClientWithTokenOrPassword(2);
     int count = 0;
     String title = null;
-    for (Trigger t : instance.getTriggers(null, true, "title", SortOrder.ASCENDING)) {
+    for (Trigger t : instance.getTriggers(null, true, "alphabetical", SortOrder.ASCENDING)) {
       if (title != null) {
-        assertTrue(title.compareTo(t.getTitle()) < 0);
```
PierreBtz, line 320: The comparison should ignore case (seen this while running against our test instance).

PR #652, opened 2024-01-09
fixes #651: adds `custom_status_id` to `Request` object

### LINKED ISSUE #651: add missing `custom_status_id` field to Ticket
`custom_status_id` is a fairly common field in zendesk and is returned by the zendesk tickets api

timveil:
this PR fixes #651.  It adds the missing `custom_status_id` field to the Response object so that is available on Ticket.  When making this change, I encountered many files which did not adhere to the `Spotless Maven Plugin` format.  This PR also applies the `google-java-format` to all files thus resolving the spotless issues.

### DIFF
--- a/.github/workflows/ci-master.yml
+++ b/.github/workflows/ci-master.yml
@@ -18,7 +18,7 @@ jobs:
           # Disabling shallow clone is recommended for improving relevancy of reporting with SonarCloud
           fetch-depth: 0
       - name: Set up JDK 1.8
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
           distribution: 'adopt'
           java-version: 8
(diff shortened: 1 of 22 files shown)


timveil:
Any thoughts on this @duemir or @PierreBtz

PierreBtz (commented):
The change itself look good.

>  When making this change, I encountered many files which did not adhere to the Spotless Maven Plugin format.

I very much doubt that (since it would fail the build), rather that you used a jdk11/17/21 take your pick instead of java 8.
I noticed in the past that the formatter had a different output with Java 8 for some reason.
The library is still using java 8.

Would it be possible for you to add a test?

Thanks for the contribution :)

timveil:
i re-ran ` mvn spotless:apply` after switching to jdk 1.8, specifically `Java version: 1.8.0_392, vendor: Amazon.com Inc.` and it looks to have reverted most of my earlier formatting changes.

As for tests, since this change is simply adding a property I don't think it requires a full blown test, but happy to copy an example test if you have one you can point me too.  In the meantime i did add the property to both `TicketTest` and `TicketResultTest`

PR #654, opened 2024-01-09
fix for #653:  add missing `custom_ticket_statuses` api call

### LINKED ISSUE #653: add support for `custom_ticket_statuses`
add support for this api call... https://developer.zendesk.com/api-reference/ticketing/tickets/custom_ticket_statuses/

timveil:
this pr fixes #653 which adds support for the `custom_ticket_statuses` api call found here...https://developer.zendesk.com/api-reference/ticketing/tickets/custom_ticket_statuses/


src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -41,45 +41,7 @@
...
-import org.zendesk.client.v2.model.TicketResult;
-import org.zendesk.client.v2.model.TimeZone;
-import org.zendesk.client.v2.model.Topic;
-import org.zendesk.client.v2.model.Trigger;
-import org.zendesk.client.v2.model.TwitterMonitor;
-import org.zendesk.client.v2.model.User;
-import org.zendesk.client.v2.model.UserField;
-import org.zendesk.client.v2.model.UserRelatedInfo;
-import org.zendesk.client.v2.model.View;
+import org.zendesk.client.v2.model.*;
```
PierreBtz, line 44: Please do not use wildcard imports.

PierreBtz (commented):
Thanks for the contribution. Same remarks as the previous PR:
* spotless need to be run using java 8
* would you mind adding a test?

Thanks!

PierreBtz:
Sorry, missed the draft status...

src/main/java/org/zendesk/client/v2/Zendesk.java
timveil: done

timveil:
I think this is ready if you are

PierreBtz (approved):
Lgtm, it seems our test instance doesn't have custom status instance enabled so I cannot have a test running on this feature.

PR #677, opened 2024-03-12
Add proper json mapping for Trigger

jonnybbb:
The trigger category isn't properly handled as the JsonProperty annotation is missing for this camel case property name


PierreBtz (dismissed):
indeed, thanks for the fix!

PierreBtz:
Marking as breaking since a Trigger persisted with a previous version will lose those two fields when deserialised.

I saw that a test is failing, I'll try to have a look.

jonnybbb:
I fixed the test cases

PR #697, opened 2024-06-03
Add VoiceComment sub-type

besbes:
In our project we need to access data about voice calls, which are represented by Zendesk as comments with a type `VoiceComment` (I added the discriminating property in #415 ). This PR adds a `VoiceComment` subclass which exposes the `data` attribute which in turn contains some additional information about voice comments (see [Adding voice comments to tickets](https://developer.zendesk.com/documentation/ticketing/managing-tickets/adding-voice-comments-to-tickets/). In particular we need to access the `call_duration` attribute, hence this is the only one that is added at the moment.

Please let me know if you think we should add more (all?) attributes, as well as how to best test this. Thank you!


PierreBtz:
@besbes thanks for the contribution. I cannot find anywhere on the Zendesk API the exact format of a comment object. in #415 it was determined it was the same as a "normal" comment (https://github.com/cloudbees-oss/zendesk-java-client/pull/415#issuecomment-968840182), but I don't see any `data` field in the comment object. How did you determine this field? Did you find some doc or you just looked at the json?

besbes:
@PierreBtz Yes… the Zendesk docs are rather weak for this specific case. I looked at the JSON returned from the Zendesk API and there I could see the `data` field for every voice comment I was looking at. There are more attributes included in the `data` field next to `call_duration` but this one is the only one we need right now so I omitted the other ones. Let me know if you want me to add more fields (just FYI, other fields include `from` and `to` (phone numbers), `recording_url`, `call_id`, `location`, `answered_by_id` etc.).

PierreBtz:
@besbes all good. Since you tested your specific case and we have a basic test to cover the more generic comment case, I'll go ahead and merge as is.

Thanks for the contribution!

PierreBtz (approved):

PR #724, opened 2024-10-28
[feature] Add support for user profiles API

mlnm-capco:
Adds support for the following User Profile CRUD operations (documented [here](https://developer.zendesk.com/api-reference/ticketing/users/profiles_api/profiles_api/#create-or-update-profile-by-user-id))

```
GET /api/v2/users/{user_id}/profiles
GET /api/v2/user_profiles/{profile_id}
GET /api/v2/user_profiles?identifier={identifier}
PUT /api/v2/users/{user_id}/profiles?identifier={identifier}
PUT /api/v2/user_profiles?identifier={identifier}
PUT /api/v2/user_profiles/{profile_id}
DELETE /api/v2/user_profiles/{profile_id}
```

### DIFF
--- a/src/main/java/org/zendesk/client/v2/Zendesk.java
+++ b/src/main/java/org/zendesk/client/v2/Zendesk.java
@@ -1346,28 +1346,94 @@ public class Zendesk implements Closeable {
             handleStatus()));
   }
 
-  public UserProfile createUserProfile(UserProfile userProfile) {
-    String idQuery = String.format(
+  public List<UserProfile> getUserProfilesForUser(User user) {
+    return getUserProfilesForUser(user.getId());
+  }
+
+  public List<UserProfile> getUserProfilesForUser(long userId) {
+    return complete(
+        submit(
+            req(
+                "GET",
+                tmpl("/users/{user_id}/profiles")
+                    .set("user_id", userId)),
+            handleList(UserProfile.class, "profiles")));
+  }
+
+  public UserProfile getUserProfile(UserProfile userProfile) {
+    return getUserProfile(userProfile.getId());
+  }
+
+  public UserProfile getUserProfile(String userProfileId) {
+    return complete(
+        submit(
+            req(
+                "GET",
+                tmpl("/user_profiles/{profile_id}")
+                    .set("profile_id", userProfileId)),
+            handle(UserProfile.class, "profile")));
+  }
+
+  public UserProfile getUserProfilebyIdentifier(String identifier) {
+    return complete(
+        submit(
+            req(
+                "GET",
+                tmpl("/user_profiles?identifier={identifier}")
+                    .set("identifier", identifier)),
+            handle(UserProfile.class, "profile")));
+  }
+
+  public UserProfile createOrUpdateUserProfile(UserProfile userProfile) {
+    return createOrUpdateUserProfile(userProfile,
+                                     userProfile.getIdentifiers().get(0).getType(),
+                                     userProfile.getIdentifiers().get(0).getValue());
+  }
+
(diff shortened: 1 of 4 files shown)


jamesforwardnwboxed:
This would be a great thing to have, @duemir @PierreBtz would you have an opportunity to get this reviewed soon? 🙌

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1344,6 +1345,84 @@ public void changeUserPassword(User user, String oldPassword, String newPassword
...
+  public List<UserProfile> getUserProfilesForUser(long userId) {
+    return complete(
+        submit(
+            req("GET", tmpl("/users/{user_id}/profiles").set("user_id", userId)),
+            handleList(UserProfile.class, "profiles")));
+  }
+
+  public UserProfile getUserProfile(UserProfile userProfile) {
+    return getUserProfile(userProfile.getId());
+  }
```
PierreBtz, line 1361: what's the purpose of this method?

PierreBtz:
Thanks for the contribution and sorry for the delay in reviewing this. I'll take some time next week to work on this.
In the meantime, in the interest of speeding things up, I did a quick scan and didn't see any obvious blocker, I just have a question about one method. 
I'll make a more thorough review later.

I'll also unblock the CI on this PR to have a run and make sure the test are passing. I can already tell you things will most likely fail because of the formatting, you should be able to format yourself with `mvn spotless:apply`.

src/main/java/org/zendesk/client/v2/model/UserProfile.java
```diff
@@ -0,0 +1,271 @@
...
+
+  private Map<String, Object> attributes;
+  private Date createdAt;
+  private String id;
+  private List<Identifier> identifiers;
+  private String name;
+  private String source;
+  private String type;
+  private Date updatedAt;
+  private String userId;
```
PierreBtz, line 21: looks odd to use a`String` here where the API (`public List<UserProfile> getUserProfilesForUser(long userId) {`) works with a `long`.

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1344,6 +1345,84 @@ public void changeUserPassword(User user, String oldPassword, String newPassword
...
+
+  public UserProfile createOrUpdateUserProfile(UserProfile userProfile) {
+    return createOrUpdateUserProfile(
+        userProfile,
+        userProfile.getIdentifiers().get(0).getType(),
+        userProfile.getIdentifiers().get(0).getValue());
+  }
+
+  public UserProfile createOrUpdateUserProfile(
+      UserProfile userProfile, String identifierType, String identifierValue) {
```
PierreBtz, line 1387: what about aligning with the `UserProfile` object and use an `Identifier` object here?

src/main/java/org/zendesk/client/v2/Zendesk.java
```diff
@@ -1344,6 +1345,84 @@ public void changeUserPassword(User user, String oldPassword, String newPassword
...
+  }
+
+  public UserProfile getUserProfile(String userProfileId) {
+    return complete(
+        submit(
+            req("GET", tmpl("/user_profiles/{profile_id}").set("profile_id", userProfileId)),
+            handle(UserProfile.class, "profile")));
+  }
+
+  public UserProfile getUserProfilebyIdentifier(String identifier) {
```
PierreBtz, line 1370: What about renaming the variable to `queryIdentifier` since this is not an identifier we are passing here, but an identifier query?

PierreBtz (changes requested):
Thanks for the contribution!
I found minor quirks that I think we should address prior to merging, but I'm ok overall with the change.

PR #817, opened 2026-04-27
Fix comment parsing without an explicit `type` present

aleksei-averchenko-wise:
The `type` field is not mandatory for comments as per https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_comments/#json-format

Normally, that doesn't matter for parsing because the type is always there - except when dry running a macro: https://developer.zendesk.com/api-reference/ticketing/business-rules/macros/#show-ticket-after-changes

I don't know why, but I did encounter this problem in production after having updated from a very ancient version that didn't use to annotate `Comment` with `@JsonTypeInfo`.

This PR fixes the problem by adding `defaultImpl = Comment.class` to the annotation, which makes Jackson use the base class by default when no type is given. I also added unit tests to verify this behaviour.


PierreBtz (approved):
lgtm, thanks!

